PR #9, opened 2018-10-19
Refactor the button and link stories

rogerfranklin2:
- Separated the stories into seperate files. I think this will make it easier to find/update when we have much more stories/scenarios.
- Used a list of types for the button so you can just select different types rather than a separate story for each usecase
screen shot 2018-10-19 at 4 32 02 pm
- Added a decorator in the config so that it adds a theme and theme provider to each component rather than having to duplicate it in the different story files. You _should_ be able to override the default theme if you need to.


rogerfranklin2:
Yeah that sounds reasonable to me. Lets see what others think

PR #12, opened 2018-10-23
Add isFullWidth and size options to Button

ldabiralai:


rogerfranklin2:
https://deploy-preview-12--eurosport.netlify.com/

Looks like netlify problems?

ldabiralai:
Try: https://deploy-preview-12--es-web-toolkit.netlify.com as we changed the netlify app name

bdo:
👍 to @rogerfranklin2's suggestion

src/atoms/Button/index.js
```diff
@@ -44,10 +44,14 @@ const Button = styled(Link)`
 
 Button.defaultProps = {
   type: 'primary',
+  size: 'default',
+  isFullWidth: false,
```
JoffreyMaheo: I think the button can take by default the full width, in fact in our react page, the button always take the full width, is it the same for you ?

bdo: How about `isAutoWidth` which defaults to false? Then we only have to specify it in the Footer and by default it’s full width.

ldabiralai:
Closing this as we agreed to handle the sizing of the button within the applications until the footer moves into here.

PR #20, opened 2018-11-05
Card component

igilfillan:
Basic Content Card plus all the components it consumes
Hover state not yet implemented


src/elements/Icon/index.test.js
```diff
@@ -0,0 +1,41 @@
...
...D0ibm9uZSIgZmlsbFJ1bGU9ImV2ZW5vZGQiPg0KICAgICAgPGcgdHJhbnNmb3JtPSJ0cmFuc2xhdGUoMTUgLjEpIj4NCiAgICAgICAgPG1hc2sgaWQ9ImIiIGZpbGw9IiNmZmYiPg0KICAgICAgICAgIDxwYXRoIGlkPSJhIiBkPSJNNiAxNC44OVYwSDB2MTQuODl6IiAvPg0KICAgICAgICA8L21hc2s+DQogICAgICAgIDxwYXRoDQogICAgICAgICAgZD0iTTUuOTk5LjQ4M0EuNDkuNDkgMCAwIDAgNS41MDMgMEguNDk3QS40OTIuNDkyIDAgMCAwIDAgLjQ4NWwuMDAxIDIuMDk2YzAgLjI4OC4xODUuNDk2LjQ3OS40OTZoMi4xMzhsLjAwMyAxMS4zM2EuNDcuNDcgMCAwIDAgLjQ3Mi40ODNoMi40MzVBLjQ3LjQ3IDAgMCAwIDYgMTQuNDA2TDUuOTk5LjQ4M3oiDQogICAgICAgICAgZmlsbD0iI0I5QkJDQSINCiAgICAgICAgICBtYXNrPSJ1cmwoI2IpIg0KICAgICAgICAvPg0KICAgICAgPC9nPg0KICAgICAgPHBhdGgNCiAgICAgICAgZD0iTTYuNzg3IDQuMzcxSDMuNTFWLjYzMWMwLS4yNjcuMjE2LS40ODQuNDgtLjQ4NGg4LjU0NWMuMjY1IDAgLjQ4LjIxOC40OC40ODV2Mi4wOTdjMCAuMjg4LS4xNzkuNDk1LS40NjQuNDk1SDYuNzg3djEuMTQ3ek0zLjM5NCAxNS4wNDNsMy4zOTMtMy4wNzdoNS43NTVjLjI3NiAwIC40NzYuMjM2LjQ3Ni40OHYyLjExM2MwIC4yNjctLjE5Mi40ODQtLjQ1Ny40ODRIMy4zOTR6Ig0KICAgICAgICBmaWxsPSIjQjlCQkNBIg0KICAgICAgLz4NCiAgICAgIDxwYXRoDQogICAgICAgIGZpbGw9IiNCOUJCQ0EiDQogICAgICAgIGQ9Ik02Ljc4NyA5LjE5Mmw0LjM0Ny0xLjg2M0g2Ljc4N1Y0LjM3MWwtMi41MiAyLjk1OEgwbDIuNzI1IDEuODYzLTIuMzQgMi43NzQgMy45MDgtMS43MDQgMi40OTQgMS43MDR6Ig0KICAgICAgLz4NCiAgICA8L2c+DQogIDwvc3ZnPg=='
```
davidgarry: It should be better to get the svg content from the map, no ?

src/components/Card/ContentCard.js
```diff
@@ -0,0 +1,95 @@
...
+ContentCard.propTypes = {
+  card: PropTypes.shape({
+    img: PropTypes.string.isRequired,
+    category: PropTypes.string.isRequired,
+    title: PropTypes.string.isRequired,
+    description: PropTypes.string,
+    timestamp: PropTypes.string,
+    channel: PropTypes.string,
+  }).isRequired,
+  type: PropTypes.string.isRequired,
```
davidgarry: I think you can use .oneOf for the type property with the different values (live, vod, article)

src/components/Card/ContentCard.test.js
```diff
@@ -0,0 +1,29 @@
...
+it('renders a ContentCard with play icon', () => {
+  const wrapper = shallow(<ContentCard card={cardData} type="vod" />);
+  expect(wrapper.find(StyledPlayIcon).length).toEqual(1);
+});
+
+it('renders a ContentCard with live label', () => {
+  const wrapper = shallow(<ContentCard card={cardData} type="live" />);
+  expect(wrapper.find(StyledLiveLabel).length).toEqual(1);
+  expect(wrapper.find(StyledPlayIcon).length).toEqual(1);
+});
```
davidgarry: Maybe add a test about the channel icon ?

src/elements/Icon/index.test.js
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import { createMatchers } from 'jest-emotion';
```
rogerfranklin2: This should be addeded to the config so all tests can use the matchers not just Icon? Seems a waste to have to do this in everytest

src/elements/Text/index.test.js
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
```
rogerfranklin2: Should have a test checking that we apply `{...props }` on the component

src/elements/UppercaseHeading.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import styled from 'react-emotion';
+
+import Heading from './Heading';
+import { turquoiseBlue } from '../colors';
+
+const StyledHeading = styled(Heading)`
+  text-transform: uppercase;
+  color: ${turquoiseBlue};
+  font-size: 0.8rem;
```
rogerfranklin2: How does a REM work when you are in a client app. Doesn't this now require everyone to set the same font size in their app? If so is this documented as a requirement of the toolkit

src/components/Card/ContentCard.js
```diff
@@ -0,0 +1,95 @@
...
+  color: ${colors.whiteLilac};
+`;
+
+export const StyledLiveLabel = styled.div`
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  padding: 0.2em 0.4em;
+  border-radius: 2px;
+  background-color: #e00034;
```
rogerfranklin2: This should be a var

src/components/Card/ContentCard.js
```diff
@@ -0,0 +1,95 @@
...
+
+const TimeStamp = styled(Text)`
+  padding: 8px 0;
+`;
+
+const StyledCardFooter = styled(Footer)`
+  * :nth-child(2) {
+    margin-left: 1em;
+    padding-left: 1em;
+    border-left: 1px solid grey;
```
rogerfranklin2: Grey. Is this the correct grey? seems lucky that it matches the HTML grey

src/components/Card/base/Content.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import styled from 'react-emotion';
+import PropTypes from 'prop-types';
+
+const StyledContent = styled('div')`
+  background-color: #10121e;
```
rogerfranklin2: colour var

src/components/Card/base/Footer.js
```diff
@@ -0,0 +1,19 @@
+import PropTypes from 'prop-types';
+import styled from 'react-emotion';
+
+const Footer = styled('div')`
+  display: flex;
+  align-items: center;
+  margin-top: 15px;
+  border-top: 1px solid grey;
```
rogerfranklin2: Grey again

src/components/Card/base/Footer.js
```diff
@@ -0,0 +1,19 @@
...
+  color: #9fa1ae;
```
rogerfranklin2: Colour

rogerfranklin2:
I wonder how much value having snapshots like this bring:

https://github.com/EurosportDigital/web-toolkit/blob/481f21b708a50a503f51b97c288adea1665f5080/src/components/Card/base/__snapshots__/Footer.test.js.snap

When it comes to a diff and an update to the component it will just be a hash change. Would it be better to setup the snapshot to show the content of the css so we can see what changes? Doesn't need to be done here just a thought

src/components/Card/ContentCard.js
```diff
@@ -0,0 +1,95 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'react-emotion';
+import { Card, Header, Content, Footer } from './base';
+import UppercaseHeading from '../../elements/UppercaseHeading';
+import Text from '../../elements/Text';
+import Icon from '../../elements/Icon';
+import { colors } from '../..';
+
+const ContentTitle = styled(Text)`
```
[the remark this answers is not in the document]
igilfillan: done

src/components/Card/ContentCard.js
```diff
@@ -0,0 +1,95 @@
...
+  position: absolute;
+  bottom: 17px;
+  left: 16px;
+`;
+
+const StyledImage = styled('img')`
+  width: 100%;
+`;
+
+const TimeStamp = styled(Text)`
```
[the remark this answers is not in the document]
igilfillan: done

src/components/Card/ContentCard.js
```diff
@@ -0,0 +1,95 @@
...
+`;
+
+const ContentCard = ({ card, type }) => {
+  const { img, category, title, description, timestamp, channel } = card;
+  const isLive = type === 'live';
+  const isPlayable = isLive || type === 'vod';
+
+  return (
+    <Card>
+      <Header style={{ position: 'relative' }}>
```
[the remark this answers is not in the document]
igilfillan: removed

src/elements/Icon/index.test.js
igilfillan: done

src/components/Card/ContentCard.js
igilfillan: done

src/components/Card/ContentCard.test.js
igilfillan: done

src/components/Card/base/Card.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'react-emotion';
+
+const StyledCard = styled('div')`
+  border-radius: 0.28571429rem;
+  box-shadow: 0 1px 3px 0 #d4d4d5, 0 0 0 1px #d4d4d5;
```
JoffreyMaheo: Adding a variable for this color.

src/components/Card/base/Header.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'react-emotion';
+
+const StyledHeader = styled.div`
```
JoffreyMaheo: could you change to => styled('div')

src/elements/UppercaseHeading.js
bdo: It does require everyone to use the same document body font size indeed. From the design system I can see that the default document font size is 16px. So let's add that to the README for now until we add typography to the toolkit.

src/components/Card/base/Card.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'react-emotion';
+
+const StyledCard = styled('div')`
+  border-radius: 0.28571429rem;
```
bdo: Not sure about the mix of px and `rems` here... I'd rather we use px for this

src/elements/Text/index.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
```
bdo: What's the point of the Text component. Currently it is just a <span>. Should we not wait until there is at least some added value until we use this abstraction?

src/elements/Heading/index.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
```
bdo: Similar to text, this is not adding any value on top of HTML headings. I suggest to inline and simplify.

src/components/Card/ScheduleCard.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import Card from './base/Card';
+
+const ScheduleCard = () => <Card>Schedule</Card>;
+
+ScheduleCard.displayName = 'ScheduleCard';
```
bdo: Do we really need the displayName? Surely this should be done by the emotion babel plugin.

src/components/Card/ContentCard.js
```diff
@@ -0,0 +1,95 @@
...
+        <StyledImage src={img} alt={description} />
+        {isPlayable && <StyledPlayIcon type="play" height="60" />}
+        {isLive && <StyledLiveLabel>Live</StyledLiveLabel>}
+      </Header>
+      <Content>
+        <UppercaseHeading as="h2">{category}</UppercaseHeading>
+        <ContentTitle>{title}</ContentTitle>
+        {description && <ContentDescription>{description}</ContentDescription>}
+        <StyledCardFooter>
+          {isLive && <Icon height="15" type={channel} />}
```
bdo: Maybe we should use `onAir` here instead of `isLive` to allow non-live on air programmes to be displayed without the [• LIVE] tag?

src/components/Card/base/Header.js
bdo: not sure about this @JoffreyMaheo. I think the dot syntax looks nicer and it's the one that's used in all examples and docs from styled-components and emotion.

bdo: I'd rather convert the other `styled('div')` to `styled.div`instead.

src/components/Card/ContentCard.js
igilfillan: This is a known issue, we had a quick discussion about this yesterday, and we agreed to leave it how it is for the time being, and we'll come back to fix this.

src/components/Card/ContentCard.js
```diff
@@ -0,0 +1,96 @@
...
+  color: ${colors.whiteLilac};
+  display: block;
+`;
+
+const StyledContentDescription = styled(Text)`
+  color: ${colors.whiteLilac};
+  display: block;
+`;
+
+export const StyledLiveLabel = styled.div`
```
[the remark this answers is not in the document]
igilfillan: done

src/components/Card/base/Header.js
igilfillan: We discussed in the team and we agreed that `styled('div')` is preferred, as it can be used for both html tags and overriding other emotion components so you don't mix and match syntax.

src/elements/Text/index.js
igilfillan: removed

src/elements/Heading/index.js
igilfillan: removed

src/components/Card/ScheduleCard.js
igilfillan: The emotion babel plugin doesn't seem to be doing this at the moment. Do you know how to get this working?

src/components/Card/base/Card.js
igilfillan: done

src/components/Card/base/Header.js
bdo: That's fine, as we've agreed to do it outside of the PR.
Tracked separately as issue #21

src/components/Card/ScheduleCard.js
bdo: Have you tried the [autoLabel option](https://github.com/emotion-js/emotion/tree/master/packages/babel-plugin-emotion#autolabel) of the emotion babel plugin?

PR #22, opened 2018-11-07
Add ability to pass attributes onto Footer links

ldabiralai:
Change Footer so the consuming application can pass props onto links for analytics purposes.

We need to pass attributes onto the footer links, but this should be defined in consuming applications, and not in the toolkit.

### DIFF
--- a/src/modules/Footer/index.js
+++ b/src/modules/Footer/index.js
@@ -79,14 +79,14 @@ const StyledCopyright = styled('div')`
 
 const Footer = ({ faq, items, copyright }) => (
   <StyledWrapper>
-    <StyledButton type="secondary" href={faq.link} target="_blank">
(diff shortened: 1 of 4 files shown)


src/modules/Footer/index.js
```diff
@@ -79,14 +79,14 @@ const StyledCopyright = styled('div')`
 
 const Footer = ({ faq, items, copyright }) => (
   <StyledWrapper>
-    <StyledButton type="secondary" href={faq.link} target="_blank">
+    <StyledButton {...faq.linkProps} type="secondary" target="_blank">
```
JoffreyMaheo: I understand why you do this for the href attribute (for never have a empty href attr) but i think it could be more clearer to keep href={faq.link} and set a default props to # instead having a anchor without href attr, but it's just my opinion, what do you think about this way ?

PR #23, opened 2018-11-08
Add props to footer component

davidgarry:
Add props to the root footer element to be able to style it in our apps, as it is already done for the link component.


Xb0X:
@bdo it's done ;) you had the same comment than Roger :)

PR #25, opened 2018-11-09
Add title prop to ContentListing module

david-sautet:


src/modules/ContentListing/Grid/index.js
```diff
@@ -30,17 +45,25 @@ const StyledItem = styled.div`
...
-  <StyledGrid>
-    {React.Children.map(children, item => (
-      <StyledItem key={item.key}>{item}</StyledItem>
-    ))}
-  </StyledGrid>
+const Grid = ({ title, children }) => (
+  <div>
+    {title && <StyledTitle>{title}</StyledTitle>}
+    <StyledGrid>
+      {React.Children.map(children, item => (
```
rogerfranklin2: I know this hasn't changed but is this wrong.

I thought the first property of map was the item(or child) and second the index?

davidgarry: Hey @rogerfranklin2 , this map function is very specific to React :
https://reactjs.org/docs/react-api.html#reactchildrenmap

PR #26, opened 2018-11-09
Add Logo and Header component

leozero:
Add Logo component (element)
Add Header component (module)


src/modules/Header/index.js
```diff
@@ -0,0 +1,26 @@
...
+  ${breakpoints.large(css`
+    height: 72px;
+  `)};
+`;
+
+const Header = () => (
+  <StyledWrapper>
+    <Logo />
+  </StyledWrapper>
+);
```
ldabiralai: We need a base snapshot test for this.

src/elements/Logo/index.js
```diff
@@ -0,0 +1,18 @@
...
+  background-repeat: none;
+  ${breakpoints.medium(css`
+    width: 197px;
+    height: 24px;
+  `)};
+`;
+
+const Logo = () => <StyledLogo />;
+
+export default Logo;
```
ldabiralai: Feels like this should be a `img` with an alt attribute?

And it feels like we should just do `export default StyledLogo` rather than create a `Logo` component.

This should also have a base snapshot

rogerfranklin2: What is best practice with SVG's now? Is it background images or inline?

src/modules/Header/index.js
leozero: Snapshot added

src/elements/Logo/index.js
leozero: Snapshot added and Logo modified to be an img with alt

bdo: > What is best practice with SVG's now? Is it background images or inline?

I don't think you guys ever discussed it. Not sure what you mean by inline here (SVG or inline base64). Let's talk about in in the Slack channel.

PR #31, opened 2018-11-14
Improve SVG support (and add webpack)

bdo:
## Adding inline SVGs

With Webpack, using inline SVGs works similarly to CSS. You can **`import` a SVG right in a JavaScript module**. This tells Webpack to include that file in the bundle. Unlike CSS imports, importing a file gives you a string value. This value is the final path you can reference in your code, e.g. as the `src` attribute of an image or the `href` of a link to a PDF.

To reduce the number of requests to the server, we returns a [data URI](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs) instead of a path. 

Here is an example:

```js
import React from 'react';
import logo from '../../assets/eurosport.svg'; // Tell Webpack this JS file uses this image

console.log(logo); // data:image/svg+xml;base64,PH...==

function Header() {
  // Import result is the URL of your image
  return <img src={logo} alt="Logo" />;
}

export default Header;
```

## Adding SVGs as React components

You can also import SVGs directly as React components. You can use either of the two approaches. In your code it would look like this:

```js
import { ReactComponent as Logo } from './logo.svg';
const App = () => (
  <div>
    {/* Logo is an actual React component */}
    <Logo />
  </div>
);
```

This is handy if you don't want to load SVG as a separate file. Don't forget the curly braces in the import! The `ReactComponent` import name is special and means that you want a React component that renders an SVG, rather than its filename.


bdo:
@rogerfranklin2 Good point, will add that

src/atoms/Link/__snapshots__/index.test.js.snap
```diff
@@ -24,6 +24,6 @@ exports[`renders a Link 1`] = `
   }
 >
   Click me !
-  <Styled(Chevron) />
+  <Styled(ReactComponent) />
```
davidgarry: As discussed with @bdo it could be better to be able to keep the "Chevron" name instead of "ReactComponent". If someone find a way to do that, it would be interesting to fix that later :)

PR #32, opened 2018-11-14
Add subnavigation component

Xb0X:
Add subnavigation component (module)


src/modules/SubNavigation/index.js
```diff
@@ -0,0 +1,64 @@
...
+
+SubNavigation.defaultProps = {
+  items: [],
+};
+
+SubNavigation.propTypes = {
+  items: PropTypes.arrayOf(
+    PropTypes.shape({
+      id: PropTypes.number,
+      label: PropTypes.string,
```
bdo: label should be required

src/modules/SubNavigation/index.js
```diff
@@ -0,0 +1,64 @@
...
+);
+
+SubNavigation.defaultProps = {
+  items: [],
+};
+
+SubNavigation.propTypes = {
+  items: PropTypes.arrayOf(
+    PropTypes.shape({
+      id: PropTypes.number,
```
rogerfranklin2: id is not required but will give a console error if not provided

src/modules/SubNavigation/index.js
rogerfranklin2: Again not required? If there is no label there probably isn't any point in having a link

src/modules/SubNavigation/index.stories.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import { object } from '@storybook/addon-knobs';
+import { withInfo } from '@storybook/addon-info';
+import SubNavigation from './index';
+
+const subNavigationStories = storiesOf('SubNavigation', module);
+const items = [
+  {
+    id: 1,
```
bdo: remove all ids

src/modules/SubNavigation/index.test.js
```diff
@@ -0,0 +1,67 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import SubNavigation from '.';
+
+const items = [
+  {
+    id: 1,
```
bdo: remove all ids

src/modules/SubNavigation/index.js
```diff
@@ -0,0 +1,64 @@
...
+
+  :hover {
+    color: white;
+  }
+`;
+
+const SubNavigation = ({ items, ...props }) => (
+  <StyledWrapper {...props}>
+    <StyledItems>
+      {items.map(item => (
```
bdo: Use destructuring:
```javascript
  {items.map(({ label, linkProps: { href, ...otherLinkProps } })=> (
```

src/modules/SubNavigation/index.js
```diff
@@ -0,0 +1,64 @@
...
+        <StyledItem key={item.id}>
+          <StyledLink {...item.linkProps}>{item.label}</StyledLink>
```
bdo: Adding the `href` explicitly has the advantage that the reader sees where it's used (it's referenced in `PropTypes` below).
```javascript
  <StyledLink href={href} {...otherLinkProps}>{label}</StyledLink>
```

src/modules/SubNavigation/index.js
```diff
@@ -0,0 +1,64 @@
...
+    </StyledItems>
+  </StyledWrapper>
+);
+
+SubNavigation.defaultProps = {
+  items: [],
+};
+
+SubNavigation.propTypes = {
+  items: PropTypes.arrayOf(
```
bdo: `items` should be required as it probably does not make sense for a submenu to have no items?

src/modules/SubNavigation/index.js
```diff
@@ -0,0 +1,64 @@
...
+      {items.map(item => (
+        <StyledItem key={item.id}>
+          <StyledLink {...item.linkProps}>{item.label}</StyledLink>
+        </StyledItem>
+      ))}
+    </StyledItems>
+  </StyledWrapper>
+);
+
+SubNavigation.defaultProps = {
```
bdo: Remove `defaultProps` if you agree with my next comment 😉

PR #33, opened 2018-11-15
Add Hero module

JoffreyMaheo:
Add viewport addons for storybook
Add font in theme file (typography) and replace it where it was used


src/modules/Hero/index.js
```diff
@@ -0,0 +1,76 @@
...
+
+const Hero = ({ title, picture }) => (
+  <StyledImage picture={picture}>
+    <StyledTitle>{title}</StyledTitle>
+  </StyledImage>
+);
+
+Hero.propTypes = {
+  title: PropTypes.string.isRequired,
+  picture: PropTypes.string.isRequired,
```
ldabiralai: Should this be `img` to be inline with the prop that's passed to `<Card />`?

JoffreyMaheo: Yes we have speaking about this point with Leopold, and we decide to name that props, picture, because we think it's a better name for images than img, maybe we can do a poll to decide the name we have to use for images ?

src/modules/Hero/index.js
```diff
@@ -0,0 +1,76 @@
...
+  font-size: 32px;
+  line-height: 40px;
+  color: ${colors.white};
+  z-index: 3;
+  margin-bottom: 137px;
+  word-break: break-word;
+  ${breakpoints.small(css`
+    font-size: 36px;
+    line-height: 44px;
+    max-width: 515px;
```
bdo: magic number here? maybe grid could help?

src/modules/Hero/index.js
```diff
@@ -0,0 +1,76 @@
...
+  `)};
+  ${breakpoints.wide(css`
+    font-size: 48px;
+    line-height: 56px;
+    max-width: 714px;
```
bdo: same here

src/modules/Hero/index.js
ldabiralai: I don't mind too much about the name, I think you could use either way. It's more that if we change to picture it'll now be a breaking change, especially as we are using the Card component in production.

src/modules/Hero/index.js
```diff
@@ -0,0 +1,76 @@
...
+  `)};
+`;
+
+const Hero = ({ title, picture }) => (
```
ldabiralai: We need to pass other props down to styled image. This makes it easy to pass down `data-test` attributes.

> Components should spread props onto the top level component, for example: const Comp = ({...props}) => <div {...props} />, this should have a test

src/elements/Grid/index.stories.js
```diff
@@ -37,7 +37,9 @@ gridStories
           medium={select('medium', ['1', '2', '3', '4', '5', '6', '7', '8', '9', '10', '11', '12'], '6')}
           large={select('large', ['1', '2', '3', '4', '5', '6', '7', '8', '9', '10', '11', '12'], '6')}
           wide={select('wide', ['1', '2', '3', '4', '5', '6', '7', '8', '9', '10', '11', '12'], '3')}
-          mediumOffset={select('medium offset', ['1', '2', '3', '4', '5', '6'], '0')}
+          tinyOffset={select('medium offset', ['1', '2', '3', '4', '5', '6'], '0')}
+          smallOffset={select('medium offset', ['1', '2', '3', '4', '5', '6'], '0')}
```
ldabiralai: Shouldn't this be tiny/small offset

david-sautet: done ;-)

src/modules/Hero/index.js
```diff
@@ -0,0 +1,105 @@
...
+    height: 630px;
+    clip-path: polygon(0 0, 100% 0%, 100% calc(100% - 50px), 0% 100%);
+  `)};
+`;
+
+const StyledRow = styled(Row)`
+  z-index: 2;
+`;
+
+const StyledTitle = styled.h1`
```
bdo: Should we not use typography here as well?

src/modules/ContentListing/Grid/index.js
```diff
@@ -15,7 +15,6 @@ const StyledGrid = styled.div`
 
 const StyledTitle = styled.h3`
```
bdo: While you're at it couldn't you *upgrade* the Grid h3 to use typography's `<H3 />`?

PR #34, opened 2018-11-16
Add component ContentArticle

david-sautet:


src/components/ContentArticle/index.js
```diff
@@ -0,0 +1,74 @@
...
+  `)};
+
+  ${breakpoints.wide(css`
+    padding: 0 584px 0 256px;
+  `)};
+`;
+
+export const StyledTeaser = styled.h2`
+  font-size: 16px;
+  font-weight: 600;
```
bdo: I don't think we have any font at `font-weight: 600`, do we?

src/components/ContentArticle/index.js
```diff
@@ -0,0 +1,74 @@
...
+import PropTypes from 'prop-types';
+import styled, { css } from 'react-emotion';
+import { breakpoints, colors } from '../../index';
+
+const StyledWrapper = styled.div`
+  padding: 0 40px;
+  color: ${colors.mirage};
+
+  ${breakpoints.medium(css`
+    padding: 0 153px;
```
bdo: `153px` looks suspiciously like a magic number. Isn't there a way to make the calculation clearer?

src/components/ContentArticle/index.js
```diff
@@ -0,0 +1,74 @@
...
+  `)};
+
+  ${breakpoints.wide(css`
+    padding: 0 584px 0 256px;
```
bdo: magic numbers again?

src/components/ContentArticle/index.js
```diff
@@ -0,0 +1,74 @@
...
+};
+
+ContentArticle.defaultProps = {
+  teaser: '',
+  paragraphs: [],
+};
+
+ContentArticle.propTypes = {
+  teaser: PropTypes.string,
+  paragraphs: PropTypes.arrayOf(
```
bdo: Why not just have paragraphs as array of strings and use the array index as the `key`? Sounds simpler...

src/components/ContentArticle/index.stories.js
```diff
@@ -0,0 +1,61 @@
...
+import { ContentArticle } from '../..';
+
+const article = storiesOf('ContentArticle', module);
+
+const data = {
+  teaser:
+    'Liverpool manager Jurgen Klopp described the pre-match praise he received from Napoli boss Carlo Ancelotti as tactics before the teams play each other in the League.',
+  paragraphs: [
+    {
+      position: 0,
```
bdo: I don't think we need the `position` there, surely the array index should suffice?

src/components/ContentArticle/index.test.js
```diff
@@ -0,0 +1,68 @@
+import { shallow } from 'enzyme/build';
+import React from 'react';
+import ContentArticle, { StyledParagraphs, StyledTeaser } from './index';
+
+const data = {
+  teaser:
+    'Liverpool manager Jurgen Klopp described the pre-match praise he received from Napoli boss Carlo Ancelotti as tactics before the teams play each other in the League.',
+  paragraphs: [
+    {
+      position: 0,
```
bdo: consider dropping `position`?

src/components/ContentArticle/index.js
david-sautet: That's what I wanted to do but Lint doesn't allow to use the array index

src/components/ContentArticle/index.js
davidgarry: For this, I think we should have a discussion about the grid. In the design, we see clearly that the text is contained between the columns 3 and 7.

src/components/ContentArticle/index.js
davidgarry: For this, I think we should have a discussion about the grid. In the design, we see clearly that the text is contained between the columns 3 and 7.

src/components/ContentArticle/index.stories.js
david-sautet: Same as before :-)

src/components/ContentArticle/index.test.js
david-sautet: Same as before :-)

src/components/ContentArticle/index.js
bdo: Why don't we add the grid now if that's useful?

src/components/ContentArticle/index.js
bdo: If you look at the instructions for the [no-array-index-key rule] (https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-array-index-key.md#prevent-usage-of-array-index-in-keys-reactno-array-index-key), you'll see why they want to enforce the rule. In our case it does not apply as the paragraphs are not going to be reordered dynamically. So you can disable the rule locally.

src/components/ContentArticle/index.js
bdo: No; if that's in the design, you need to talk to Alex and Jon because we only have bold and normal fonts.

david-sautet: Here the response of Jon and Alex : 

> we’d only use SemiBold - but to be honest, by using the SemiBold web font, you might have to set the weight to 400 (not 600) so it gives you that font’s ‘default thickness’
> if you look here: https://sport-design-system.netlify.com/style-typography you’ll see how I had to set anything <strong> / <bold> to use the InterUISemiBold font, and set font-weight to 400 in the CSS, otherwise the browser was thickening the original font

So for the moment we will wait for the "Typography" story to be done to change this.

src/components/ContentArticle/index.js
david-sautet: A story as been created in order to add the grid

src/components/ContentArticle/index.js
david-sautet: This is now ok

src/components/ContentArticle/index.stories.js
david-sautet: I now use the index and thus change the type of paragraphs to array of strings

src/components/ContentArticle/index.test.js
david-sautet: I now use the index and thus change the type of paragraphs to array of strings

src/components/ContentArticle/index.js
bdo: Please change it to **bold** for now as I have asked Alex to change it in the SDS

david-sautet: it's now ok

PR #36, opened 2018-11-19
Spread props passed to Header

ldabiralai:
This is to satisfy: https://github.com/EurosportDigital/web-toolkit/blob/master/CONTRIBUTING.md#requirements

> Components should spread props onto the top level component, for example: const Comp = ({...props}) => <div {...props} />, this should have a test

### DIFF
--- a/src/modules/Header/index.js
+++ b/src/modules/Header/index.js
@@ -17,8 +17,8 @@ const StyledWrapper = styled.div`
   `)};
 `;
 
-const Header = () => (
-  <StyledWrapper>
+const Header = props => (
+  <StyledWrapper {...props}>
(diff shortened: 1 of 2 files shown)


src/modules/Header/index.js
```diff
@@ -17,8 +17,8 @@ const StyledWrapper = styled.div`
   `)};
 `;
 
-const Header = () => (
-  <StyledWrapper>
+const Header = ({ ...props }) => (
```
bdo: Did you not mean the following?
```suggestion
const Header = props => (
```

ldabiralai: Good point.. 😄

PR #37, opened 2018-11-19
manage link if url available in card content

Xb0X:


src/components/Card/ContentCard.js
```diff
@@ -163,7 +164,7 @@ const ContentCard = ({ card, type, ...props }) => {
 ContentCard.propTypes = {
```
ldabiralai: How can we make it clear that `linkComponent` is a prop here (and what that prop can/should be)

bdo: Good point, this need to be documented in the README.md

bdo: To clarify the purpose, it allows you to override the link behaviour in your app when you need to use a router link.

For instance if you wanted to use React Router for internal links, you could use something like think:
```js
import React from 'react';
import { Link as EurosportWebToolkitLink } from '@eurosport/web-toolkit';
import { Link as ReactRouterLink } from 'react-router-dom';

const isExternalUrl = url => url && url.match(/^(https?|mailto|tel):/);

EurosportWebToolkitLink.defaultProps.linkComponent = ({ href, children, ...props }) => {
  if (isExternalUrl(href)) {
    return (
      <a {...props} href={href}>
        {children}
      </a>
    );
  }
  return (
    <ReactRouterLink {...props} to={href}>
      {children}
    </ReactRouterLink>
  );
};

```

ldabiralai: @bdo, I understand the purpose, but it's very hidden away. My concern is that it's not discoverable, how can we make it obvious?

Is the prop listed in the storybook documentation?

PR #40, opened 2018-11-20
Fix callsign matching in channel icon

ldabiralai:


src/elements/ChannelIcon/index.test.js
```diff
@@ -1,6 +1,12 @@
 import React from 'react';
 import { shallow } from 'enzyme';
-import ChannelIcon, { iconMap } from '.';
+import ChannelIcon from '.';
+import E from '../../assets/channels/E.svg';
```
davidgarry: Maybe we could keep the iconMap to avoid import again here all the svg. What do you think ?

src/elements/ChannelIcon/index.js
```diff
@@ -7,41 +7,56 @@ import E2GR from '../../assets/channels/E2GR.svg';
 import E2NO from '../../assets/channels/E2NO.svg';
 import E2RUG from '../../assets/channels/E2RUG.svg';
```
davidgarry: Maybe we can keep here the iconMap for the tests ? (see other comment)

src/elements/ChannelIcon/index.js
```diff
@@ -7,41 +7,56 @@ import E2GR from '../../assets/channels/E2GR.svg';
...
 
-export const iconMap = {
-  E: {
+const getChannelIcon = callsign => {
```
bdo: I'm thinking something like this would actually read better:
```suggestion
import * as channels from '../../assets/channels';

const baseChannel = callSign => callSign.match(/E\d?/)[0];

const getChannelIcon = callSign => channels[callSign] || channels[baseChannel(callSign)] || channels.E;
```

bdo: And I'll move all the data into `../../assets/channels/index.js`

like this:
```js
import e from './E.svg';
import e1 from './E1.svg';
import e2 from './E2.svg';
import e2GR from './E2GR.svg';
import e2NO from './E2NO.svg';
import e2RUG from './E2RUG.svg';

export const E = {
  src: e,
  altText: 'Eurosport',
  widthRatio: 0.87,
};

export const E1 = {
  src: e1,
  altText: 'Eurosport 1',
  widthRatio: 1.42,
};

export const E2 = {
  src: e2,
  altText: 'Eurosport 2',
  widthRatio: 1.66,
};

export const E2GR = {
  src: e2GR,
  altText: 'Eurosport 2',
  widthRatio: 3.5,
};

export const E2NO = {
  src: e2NO,
  altText: 'Eurosport 2',
  widthRatio: 2.06,
};

export const E2RUG = {
  src: e2RUG,
  altText: 'Eurosport 2',
  widthRatio: 4,
};
```

src/elements/ChannelIcon/index.test.js
```diff
@@ -1,6 +1,12 @@
...
+import E1 from '../../assets/channels/E1.svg';
+import E2 from '../../assets/channels/E2.svg';
+import E2GR from '../../assets/channels/E2GR.svg';
+import E2NO from '../../assets/channels/E2NO.svg';
+import E2RUG from '../../assets/channels/E2RUG.svg';
```
bdo: Then we would have to import like this:
```js
import { E, E1, E2, E2GR, E2RUG, E2NO } from '../../assets/channels';
```

and replace occurences of `E`, `E1`, ... with `E.src`

PR #35, opened 2018-11-16
Add Author Bloc, and rename button/link stories to index.stories.js

JoffreyMaheo:

PR #47, opened 2018-11-22
Add Author Bloc

bdo:
Simplify PR #35 as it was just too old 🌵


bdo:
Same content already approved in #35 by @ldabiralai and myself - I just removed things that had nothing to do with PR such as renaming button/link stories to index.stories.js. These caused problems when attempting to rebase or merge and it was simpler to restart from scratch.

PR #51, opened 2018-11-26
Ewp 307 article module

davidgarry:


src/modules/Article/index.js
```diff
@@ -0,0 +1,36 @@
...
+
+const StyledArticleContent = styled(ArticleContent)`
+  margin: 37px 0;
+  ${breakpoints.medium(css`
+    margin: 47px 0;
+  `)};
+`;
+
+const Article = ({ data: { title, img, author, time, teaser, paragraphs }, ...props }) => (
+  <div {...props}>
```
igilfillan: Using the `<article>` tag would give better semantics

david-sautet: You're right :-)

PR #52, opened 2018-11-26
Fix hero background

davidgarry:
Update hero/author colors and opacity


src/modules/Hero/index.js
```diff
@@ -66,14 +78,15 @@ const StyledAuthor = styled(Author)`
 `;
 
 const Hero = ({ title, img, author, time, ...props }) => (
-  <StyledImage img={img} {...props}>
+  <StyledWrapper>
+    <StyledImage img={img} {...props} />
```
ldabiralai: > Components should spread props onto the top level component, for example: const Comp = ({...props}) => <div {...props} />, this should have a test

Any reason this is on StyledImage? What props are we expecting there?

I think this should live on StyledWrapper, and any props to go on the image can be passed in as imageProps and then spread.

davidgarry: Well spotted @ldabiralai ! And I will add a test to cover that ;)

PR #53, opened 2018-11-27
Add schedule card and extract Card Details component for reuse between Content card and Schedule card

igilfillan:
As of this PR we now will show the channel icon if a channel is passed to either card, before this would only show if the card type was live. This will be a problem if you are passing the channel into the content card when you don't want the icon shown. So now it's the consumers responsibility to not pass that if it's not wanted.

### DIFF
--- a/src/components/Card/CardDetails.js
+++ b/src/components/Card/CardDetails.js
@@ -1,0 +1,106 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'react-emotion';
+import { rgba } from 'polished';
+import ChannelIcon from '../../elements/ChannelIcon';
+import * as colors from '../../colors';
+
+const StyledContent = styled.div`
+  padding: 16px 16px 8px;
+  line-height: 1.3;
+  flex: 1 auto;
+  display: flex;
+  flex-direction: column;
+`;
+
+const StyledCategory = styled.div`
+  font-size: 12px;
+  font-weight: bold;
+  text-transform: uppercase;
+  color: ${colors.turquoiseBlue};
+  letter-spacing: 1px;
+  margin: 0 0 10px;
+`;
+
+const footerBorder = rgba(colors.whiteLilac, 0.15);
+const StyledFooter = styled.div`
+  flex: 1 auto;
+  display: flex;
+  align-items: flex-end;
+  margin-top: 20px;
+  font-size: 14px;
+`;
+
+const StyledDetails = styled.div`
+  display: flex;
+  align-items: center;
+  width: 100%;
+  border-top: 1px solid ${footerBorder};
+
+  * :nth-child(2) {
+    margin-left: 10px;
+    padding-left: 10px;
+    border-left: 1px solid ${footerBorder};
+    padding: 8px;
+  }
+`;
+
+const StyledTitle = styled.div`
+  font-size: 14px;
+  font-weight: bold;
+  color: ${colors.whiteLilac};
+  margin: 0 0 10px;
+`;
+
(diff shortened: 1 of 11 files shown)


src/components/Card/ScheduleCard.js
```diff
@@ -0,0 +1,82 @@
...
+  display: block;
+  position: absolute;
+  opacity: 0.35;
+
+  :before {
+    content: '';
+    position: absolute;
+    z-index: 1;
+    height: 100%;
+    width: 101%;
```
davidgarry: Why 101% ?

src/components/Card/ScheduleCard.js
```diff
@@ -0,0 +1,82 @@
...
+
+const ScheduleCard = ({ card, ...props }) => (
+  <StyledWrapper {...props}>
+    <StyledImage img={card.img} />
+    <StyledCardDetails card={card} />
+  </StyledWrapper>
+);
+
+ScheduleCard.propTypes = {
+  card: PropTypes.shape({
```
davidgarry: The `card` prop should be required no ?

src/components/Card/ScheduleCard.js
```diff
@@ -0,0 +1,82 @@
...
+    content: '';
+    position: absolute;
+    z-index: 1;
+    height: 100%;
+    width: 101%;
+    background: linear-gradient(
+      180deg,
+      rgba(21, 23, 38, 0) 0%,
+      rgba(21, 23, 38, 0.3) 18.66%,
+      rgba(21, 23, 38, 0.85) 100%
```
davidgarry: Can you export this color in colors.js ?

src/components/Card/ContentCard.js
```diff
@@ -126,17 +72,17 @@ export const StyledLiveLabel = styled.div`
...
 
-const StyledTimeStamp = styled.div`
-  padding: 8px 0;
-  margin: 0;
-  color: ${colors.manatee};
-  font-size: 12px;
-`;
-
 const ContentCard = ({ card, type, ...props }) => {
-  const { img, url, category, title, description, timestamp, channel, liveLabel } = card;
```
davidgarry: Maybe you can use this syntax : 
`const { img, url, liveLabel, ...cardDetailsProps } = card;`

And remove the definition of `cardDetailsProps` below.

src/components/Card/CardDetails.js
```diff
@@ -0,0 +1,106 @@
...
+          {channel && <ChannelIcon height={15} type={channel} />}
+          <StyledTimeStamp>{timestamp}</StyledTimeStamp>
+        </StyledDetails>
+      </StyledFooter>
+    </StyledContent>
+  );
+};
+
+CardDetails.propTypes = {
+  card: PropTypes.shape({
```
davidgarry: `card` should be required

src/components/Card/ScheduleCard.js
ldabiralai: @JoffreyMaheo had the same comment a while ago. Without that the gradient doesn't seem to fill the whole width of the card.

image

src/components/Card/ScheduleCard.js
ldabiralai: Good point, done!

src/components/Card/CardDetails.js
ldabiralai: Done!

src/components/Card/ScheduleCard.js
ldabiralai: Done!

PR #54, opened 2018-11-27
Little fix on Author component if the img is an empty string (=> disp…

david-sautet:
Little fix on Author component if the img is an empty string (=> display default img)

### DIFF
--- a/src/components/Author/index.test.js
+++ b/src/components/Author/index.test.js
@@ -14,7 +14,7 @@ describe('Author', () => {
   });
   it('renders Author without avatar', () => {
     const wrapper = mount(<Author name="name" time="time" />);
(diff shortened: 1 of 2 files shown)


src/components/Author/index.test.js
```diff
@@ -16,4 +16,8 @@ describe('Author', () => {
     const wrapper = mount(<Author name="name" time="time" />);
     expect(wrapper.prop('img')).toEqual(E);
   });
```
ldabiralai: Should this test be doing `expect(wrapper.find('img').prop('src')).toEqual(E);` (like the one we've added below?)

david-sautet: Yes it is the same it's just that the first one has the defaultProp equals to E.
I made the change you request anyway as it will more clearer :-)

PR #60, opened 2018-11-29
Export hover styles so that the Card can control when they should be added, and prefer loading state to hover state

ldabiralai:
As of https://github.com/EurosportDigital/web-toolkit/pull/59 the hover state on the live Card play button is broken.

This fixes it by exporting the hover styles internally so that the Card no longer has to have knowledge of the `PlayIcon` internals to add a hover state meaning this shouldn't happen again.

I initially added a `withHoverState` prop, so that consumers could add the hover state if they wanted it, but I've removed it as I think it's safe to say that it should always be there. At least it does from our teams perspective.

### DIFF
--- a/src/elements/PlayIcon/index.js
+++ b/src/elements/PlayIcon/index.js
@@ -47,7 +47,7 @@ const StyledWrapper = styled.div`
   overflow: hidden;
 
   :hover {
-    ${({ withHoverState }) => withHoverState && hoverStyles}
+    ${hoverStyles}
   }
 `;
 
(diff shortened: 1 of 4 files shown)


src/elements/PlayIcon/index.js
```diff
@@ -41,8 +29,30 @@ const StyledBorder = styled.div`
   animation: ${({ isLoading }) => isLoading && `${spin} 1s linear infinite`};
 `;
 
+export const hoverStyles = css`
+  ${StyledBorder} {
+    border-color: white;
```
david-sautet: I think you need to use ${white} instead of the hardcoded value "white"

PR #64, opened 2018-12-11
Ewp 346 fix grid gap issue

leozero:
Remove the space-between in the grid component
Add padding on column component
Add negative margin in the row sub container
Recalculate the gap / margin / width to match the new negative margin + column padding system

### DIFF
--- a/src/elements/Grid/index.js
+++ b/src/elements/Grid/index.js
@@ -88,7 +88,8 @@ export const getColumnStyles = (colNumber, offset, size) => {
   const columns = colNumber === 'full' ? grid[size].columns : colNumber;
   const width = getWidthString(columns, size);
   const offsetStyle = getOffset(offset, size);
-  return `${width}${offsetStyle}`;
+  const padding = `padding: 0 ${Math.floor((grid[size].gap / 2) * 100) / 100 + grid[size].type};`;
+  return `${width}${offsetStyle}${padding}`;
 };
 
 export const Container = styled.div`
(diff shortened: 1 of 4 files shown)


src/elements/Grid/index.js
```diff
@@ -177,6 +202,14 @@ export const Column = props => {
   );
 };
 
+Row.propTypes = {
+  children: PropTypes.node,
+};
+
+Row.defaultProps = {
+  children: null,
```
bdo: Why do we need to set default props? Can't it be left as `undefined`?

src/elements/Grid/index.js
```diff
@@ -111,12 +113,35 @@ export const Container = styled.div`
...
+  ${breakpoints.large(css`
+    ${getSubContainerMargin('large')};
+  `)};
+
+  ${breakpoints.wide(css`
+    ${getSubContainerMargin('wide')};
+  `)};
 `;
 
+export const Row = (props, { children }) => (
```
bdo: Did you not mean the following?
```suggestion
export const Row = ({ children, ...props }) => (
```

src/elements/Grid/index.js
```diff
@@ -177,6 +202,14 @@ export const Column = props => {
...
+  children: PropTypes.node,
+};
+
+Row.defaultProps = {
+  children: null,
+};
+
 Column.defaultProps = {
   tiny: 'full',
   small: null,
```
bdo: Similar thing here while you're at it: if a props can have a undefined value, it is easier because you don't have to default it! Unless there is a valid reason for this?

src/elements/Grid/index.js
```diff
@@ -111,12 +113,35 @@ export const Container = styled.div`
   `)};
 `;
 
-export const Row = styled(Container)`
+const getSubContainerMargin = size => `margin: 0 ${Math.floor((grid[size].gap / -2) * 100) / 100 + grid[size].type};`;
```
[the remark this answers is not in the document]
leozero: Done

src/elements/Grid/index.js
leozero: Done

src/elements/Grid/index.js
leozero: Done

src/elements/Grid/index.js
leozero: Done

src/elements/Grid/index.js
```diff
@@ -3,9 +3,10 @@ import styled, { css } from 'react-emotion';
 import PropTypes from 'prop-types';
 import * as breakpoints from '../../breakpoints';
 
+// ratio is the content width in percent plus 2 half gaps (so one gap) because of negative margins on the row
```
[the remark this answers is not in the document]
bdo: @leozero Do you think you could try @igilfillan 's suggestion? Thanks 😄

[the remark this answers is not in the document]
Xb0X: Hi @igilfillan and @bdo. After investigations it seems more complex to calculate and harder to maintain using box-sizing. Indeed, we need to recalculate all the ratios. Without sub container it would have been more simple but here it add useless complexity for us... :)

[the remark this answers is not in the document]
bdo: @Xb0X fair enough ! Thanks for trying 👍

PR #66, opened 2018-12-18
Eurweb 888 latest cards

igilfillan:
Adds new VOD Compact card
Adds CompactCard
Updates Schedule card to use CompactCard
Updates CardDetails component to accept a node to display as the icon in footer


src/components/Card/CompactCard.js
```diff
@@ -0,0 +1,84 @@
...
+  position: absolute;
+  opacity: 0.35;
+  z-index: 0;
+
+  :before {
+    content: '';
+    position: absolute;
+    z-index: 1;
+    height: 100%;
+    width: 101%;
```
JoffreyMaheo: I don't know if it could be resolve the problem, but if is, it should be a cleaner way to do like that: bottom:0; top:0; left:0; right:0;

src/components/Card/ScheduleCard.js
```diff
@@ -1,58 +1,11 @@
...
-`;
+import ChannelIcon from '../../elements/ChannelIcon';
+import CompactCard from './CompactCard';
 
 const ScheduleCard = ({ card, ...props }) => (
-  <StyledWrapper {...props}>
-    <StyledImage img={card.img} />
-    <StyledCardDetails card={card} />
-  </StyledWrapper>
+  <CompactCard card={card} icon={<ChannelIcon height={15} type={card.channel} />} {...props} />
```
JoffreyMaheo: Maybe for a code more readable, you can store <ChannelIcon /> in a const Icon ?

src/components/Card/CompactCard.js
igilfillan: I have set the width to !00% and it works now. Strange

src/components/Card/ScheduleCard.js
igilfillan: I agree it would be a bit easier to read but as the const wouldn't be reused, and it's not too difficult to read, we feel we should keep it like it is.

PR #68, opened 2019-01-03
Add colours defined in design system and removed unused colours in toolkit

ldabiralai:
Breaking change!!

- I've removed any colours that are not found in the toolkit components. Consumers should double check that any colours they need are available, and if not, comment which ones and I can add them (or add yourself).
- New colours should be present in: https://sport-design-system.netlify.com/style-colour if not we should talk to UI and confirm which colour we should be using in the component and ensure it matches one of these.

~~There are some old colours that I have left in the colors.js file, pending a conversation with UI around which colours should replace these usages.~~


src/assets/channels/E2NO.svg
```diff
@@ -1,5 +1,5 @@
 <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 47 23">
-  <g fill="#B9BBCA">
+  <g fill="#c6c7ce">
```
dstodolny: If we would use something like: https://github.com/smooth-code/svgr then we can pass these colours as props by importing them from colors.js file.

ldabiralai: I agree that it's not great to hardcode these values, but I think for now we should regard it as some tech debt as it'll change too much of the implementation for this PR.

I'll raise an issue for it, so we don't forget about this

ldabiralai: https://github.com/EurosportDigital/web-toolkit/issues/70

PR #71, opened 2019-01-07
Adding * border-box

rogerfranklin2:
This is a big breaking change if people aren't using border-box

### DIFF
--- a/src/elements/Grid/index.js
+++ b/src/elements/Grid/index.js
@@ -65,7 +65,7 @@ const SIX_OFFSETS = ['0', '1', '2', '3', '4', '5'];
 const TWELVE_OFFSETS = ['0', '1', '2', '3', '4', '5', '6', '7', '8', '9', '10', '11'];
 
 export const getWidthString = (colNumber, size) => {
-  const width = colNumber * grid[size].columnSize + (colNumber - 1) * grid[size].gap;
+  const width = colNumber * grid[size].columnSize + (colNumber - 1) * grid[size].gap + grid[size].gap;
   return `width: ${width + grid[size].type};`;
 };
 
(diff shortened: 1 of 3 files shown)


ldabiralai:
@rogerfranklin2 I think this will break the grid, as the grid is currently using `padding` between columns.

rogerfranklin2:
To be honest don't really have the time or will power to go through and add to all components individually.

 Surely this means that the grid/toolkit doesn't work correctly in your app?

https://github.com/EurosportDigital/watch-hub/blob/0ed9219c5ad22f6861f8c1a6fc869a9b1d3999fe/src/customReset.js#L27

Happy to close this PR and just keep it in our app if the components have been designed without border-box

src/elements/Grid/index.js
```diff
@@ -148,6 +148,7 @@ export const Row = ({ children, ...props }) => (
 );
 
 export const StyledColumn = styled.div`
+  box-sizing: border-box;
   ${({ tiny, small, medium, large, wide, tinyOffset, smallOffset, mediumOffset, largeOffset, wideOffset }) => css`
     ${getColumnStyles(tiny, tinyOffset, 'tiny')};
```
igilfillan: From the storybook it's very difficult to see if the grid is correct or not.  Ideally there would be a defined outline which represented the page, and we could see the columns inside that outline. Then we would be able to see if the items were butted up against the outlined box or not. So it's quite difficult to see if your changes have fixed the problem. But what I imagined is that the new grid would need a bit more rework than what you've done, I could be wrong. Using border-box, the width calculation for each column should be relatively straight forward something like `width: numberOfCols / totalNumberOfCols`. Each column would have a left and right padding, which when added together would be equivalent to the column gutters. But this would mean that the first item would have a left padding that needs to be removed, and the last column would have a right padding which needs to be removed. This could be handled by the addition of a parent container with negative left and right margins which matched the padding  value.

igilfillan: Looking into tis further, I can see how we can refactor it, however it is probably best handled in a separate ticket.

PR #72, opened 2019-01-08
Update content listing grid to use the toolkit grid

ldabiralai:


src/modules/ContentListing/Grid/__snapshots__/index.test.js.snap
```diff
@@ -1,27 +1,62 @@
...
 
 exports[`ContentListing Grid base grid 1`] = `
-<Styled(section)>
-  <Styled(div) />
-  <Styled(div)>
-    <Styled(div)
+<Styled(div)>
+  <Row>
+    <Column
+      large={null}
```
igilfillan: Do you have to pass in `null`? I thought they had set this up so that it would take the value of the previous bp if nothing was passed for the next bp.

PR #76, opened 2019-01-14
Watch bar card

Xb0X:
We start to create a component which will be use on the eurosport website.

Also we made an evolution on how to implement the SVG for channels logos. We go from  to <svg/> in order to fill it with its own color according to the use.


src/components/Card/WatchbarCard.js
```diff
@@ -0,0 +1,163 @@
...
+  flex-direction: column;
+  justify-content: space-between;
+  background: ${colors.actionTwoDarkBase};
+  font-size: 13px;
+`;
+
+const StyledFooter = styled.div`
+  color: rgba(251, 251, 255, 0.8);
+  font-size: 11px;
+  border-top: 1px solid rgba(242, 243, 245, 0.4);
```
ldabiralai: We should only be using colours that are present in: https://sport-design-system.netlify.com/style-colour same two lines above. Could you talk to @jonwelchdiscovery to see what we should use here?

src/components/Card/WatchbarCard.js
```diff
@@ -0,0 +1,163 @@
...
+
+  return (
+    <StyledCard {...props} href={url}>
+      <StyledHeader>
+        <StyledImage src={img} />
+        <StyledChannelIcon height={15} type={card.channel} />
+        <StyledPlayIcon height={32} />
+        {isLive && <StyledLiveLabel>{liveLabel}</StyledLiveLabel>}
+      </StyledHeader>
+      <StyledContent {...props}>
```
ldabiralai: I think we should only be spreading `{...props}` on the `StyledCard`

src/components/Card/WatchbarCard.js
```diff
@@ -0,0 +1,163 @@
...
+    isLive: PropTypes.bool,
+    title: PropTypes.string.isRequired,
+    timestamp: PropTypes.string,
+    channel: PropTypes.string,
+    liveLabel: PropTypes.string,
+  }).isRequired,
+  linkComponent: PropTypes.func,
+};
+
+WatchbarCard.displayName = 'Cards.Content';
```
ldabiralai: Should this be `Cards.Watchbar`?

src/components/Card/WatchbarCard.test.js
```diff
@@ -0,0 +1,23 @@
...
+
+it('renders a WatchbarCard', () => {
+  const wrapper = shallow(<WatchbarCard card={cardData} />);
+  expect(wrapper).toMatchSnapshot();
+});
+
+it('renders a WatchbarCard with a live label', () => {
+  const wrapper = shallow(<WatchbarCard card={{ ...cardData, isLive: true }} />);
+  expect(wrapper.find(StyledLiveLabel).length).toEqual(1);
+});
```
ldabiralai: We should have a test to check for prop spreading

src/elements/ChannelIcon/getChannelIcon.js
```diff
@@ -1,40 +1,40 @@
...
   },
   E2: {
-    src: E2,
     altText: 'Eurosport 2',
     widthRatio: 1.66,
+    component: E2Component,
   },
   E: {
-    src: E,
     altText: 'Eurosport',
```
ldabiralai: I think we can remove these as I don't think they're used?

ldabiralai: (the `altText` attributes)

src/components/Card/WatchbarCard.js
Xb0X: You're right we're now using a color from the sds 👍

src/components/Card/WatchbarCard.js
Xb0X: done

src/components/Card/WatchbarCard.js
Xb0X: done

src/components/Card/WatchbarCard.test.js
Xb0X: done

src/elements/ChannelIcon/getChannelIcon.js
Xb0X: done

PR #78, opened 2019-01-23
Update install instructions

rogerfranklin2:
The toolkit doesn't work with version 10 of emotion so need to specify version 9


README.md
```diff
@@ -6,7 +6,7 @@ A collection of components to share common functionality across teams.
 
 ## How to use
 
-`npm install @eurosport/web-toolkit emotion react-emotion emotion-theming polished`
+`npm install @eurosport/web-toolkit emotion@9.2 react-emotion@9.2 emotion-theming@9.2 polished@2.3`
```
bdo: Do we want to limit it to major revisions instead of minor?
```suggestion
`npm install @eurosport/web-toolkit emotion@9 react-emotion@9 emotion-theming@9 polished@2`
```

PR #79, opened 2019-01-28
Carousel

Xb0X:
Hey there, for the need to integrate the watchbar as a react component on the legacy website, we developed here a carousel system to manage a slider view.

We build it from scratch in an iterative way to match the needs of the watchbar.

### DIFF
--- a/src/components/Carousel/index.js
+++ b/src/components/Carousel/index.js
@@ -14,7 +14,7 @@ const StyledSlidesTrack = styled.div`
   display: block;
   margin-left: auto;
   margin-right: auto;
-  width: ${props => props.nbSlides * props.slideWidth}px;
+  width: ${props => props.length * props.slideWidth}px;
   transition: left 200ms ease-out;
 `;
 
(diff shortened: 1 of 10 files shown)


src/components/Carousel/index.js
```diff
@@ -0,0 +1,226 @@
...
+const StyledChevron = styled(Chevron)`
+  height: 16px;
+  width: 11px;
+  path {
+    fill: rgba(255, 255, 255, 0.7);
+  }
+`;
+
+export default class Carousel extends React.Component {
+  constructor(props) {
```
bdo: Can be simplified by moving state and refs outside of constructor.

src/components/Carousel/index.js
```diff
@@ -0,0 +1,226 @@
...
+    super(props);
+    this.state = {
+      left: 0,
+      currentSlide: 0,
+      slideMargin: props.slideMargin,
+      trackWidth: 0,
+    };
+    this.wrapperRef = React.createRef();
+    this.slidesTrackRef = React.createRef();
+    this.calculateElementsSizes = debounce(this.calculateElementsSizes, 200);
```
bdo: No need for debounce if the objective is just to wait 200ms before calculating the element sizes.

src/components/Carousel/index.js
```diff
@@ -0,0 +1,226 @@
...
+  }
+
+  componentDidMount() {
+    this.calculateElementsSizes();
```
bdo: Calculate the element size after 200ms here

src/components/Carousel/index.js
```diff
@@ -0,0 +1,226 @@
...
+    window.addEventListener('resize', this.calculateElementsSizes.bind(this));
+  }
+
+  getSlidesInformations(children) {
+    const { slideMargin } = this.state;
+    let position = 0;
+    const slides = [];
+    Array.from(children).forEach(slide => {
```
bdo: Prefer using `reduce` instead of `forEach` here

src/components/Carousel/index.js
```diff
@@ -0,0 +1,226 @@
...
+      };
+      position += information.width;
+      slides.push(information);
+    });
+    return slides;
+  }
+
+  static getTrackWidth(slides) {
+    let width = 0;
+    slides.forEach(slide => {
```
bdo: Prefer using `reduce` instead of `forEach` here

src/components/Carousel/index.js
```diff
@@ -0,0 +1,226 @@
...
+      width += slide.width;
+    });
+    return width;
+  }
+
+  isInRange(left) {
+    const { trackWidth, wrapperWidth, slides } = this.state;
+    return !(left > 0 || left < trackWidth * -1 + wrapperWidth - slides[slides.length - 1].width);
```
bdo:
```suggestion
    return !(left > 0 || left < wrapperWidth - trackWidth - slides[slides.length - 1].width);
```

bdo: Also the `isInRange` method seems to determine whether the point x in outside the range (because of the negation). Suggest either to rename to `isOutOfRange` or remove the negation.

src/components/Carousel/index.js
```diff
@@ -0,0 +1,226 @@
...
+  }
+`;
+
+export default class Carousel extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      left: 0,
+      currentSlide: 0,
+      slideMargin: props.slideMargin,
```
bdo: Why add `slideMargin` to the state, probably better to read it from props always as it does not change?

src/components/Carousel/index.js
```diff
@@ -0,0 +1,226 @@
...
+      this.setState({
+        left,
+        currentSlide,
+      });
+    }
+  }
+
+  render() {
+    const { children } = this.props;
+    const { left, trackWidth, slideMargin } = this.state;
```
bdo: Read `slideMargin` from props here

src/components/Carousel/index.js
Xb0X: done :)

src/components/Carousel/index.js
Xb0X: done :)

src/components/Carousel/index.js
Xb0X: done :)

src/components/Carousel/index.js
Xb0X: done :)

src/components/Carousel/index.js
Xb0X: done :)

src/components/Carousel/index.js
Xb0X: done :)

src/components/Carousel/index.js
Xb0X: done :)

PR #86, opened 2019-03-05
Add quantcast and test attributes

val3ntin:
this adds:
1. quantCast label to the BurgerMenu
2. data-test attrs to the Logo / Header / BurgerMenu
3. "react-docgen" to the .babelrc so as to fix the propTypes display into the Header storybook component (related  [storybook issue](https://github.com/storybooks/storybook/issues/2708) )


val3ntin:
> Good work, but why add the docgen plugin ?

compare prop-types description:

before:
screen shot 2019-03-05 at 3 43 35 pm

after:
screen shot 2019-03-05 at 3 43 16 pm

https://github.com/storybooks/storybook/issues/2708

PR #88, opened 2019-03-06
Ewl 336 eurosport logo dimensions

val3ntin:
this PR enables similar header styling as on legacy website:
1. burger icon margins.
2. logo icon width/height


src/modules/BurgerMenu/LeftMenu/LeftMenu.js
```diff
@@ -47,9 +47,19 @@ const StyledTabLogo = styled.a`
     height: 157px;
   `)};
 
-  img {
-    width: 100%;
+  &&& ${Logo.css} {
```
arturkin: ☢︎

val3ntin: https://www.styled-components.com/docs/faqs#how-can-i-override-styles-with-higher-specificity

PR #91, opened 2019-03-15
add data tracking on watchbarCard link

Xb0X:
BugFix :
Add missing data tracking attribute on watchbar card link


src/components/Card/index.stories.js
```diff
@@ -87,7 +90,7 @@ cardStories
     'Watchbar',
     withInfo({ propTablesExclude: [Wrapper] })(() => (
       <Wrapper>
-        <Cards.Watchbar card={object('card', watchbar)} />
+        <Cards.Watchbar card={object('card', watchbar)} trackingPosition={object('card', watchbar.trackingPosition)} />
```
davidgarry: The name of the second object property is incorrect. It should be "trackingPosition" instead of "card"

Xb0X: Fixed :)

PR #94, opened 2019-03-21
Ewl 406 watchbar text overflow

arturkin:
Improve UI rendering https://eurosportdigital.atlassian.net/browse/EWL-413
Text ellipsis fix https://eurosportdigital.atlassian.net/browse/EWL-406


src/modules/Watchbar/index.js
```diff
@@ -80,8 +80,13 @@ const generateChildren = (cards, title) => {
 
 const WatchBar = ({ cards, title }) => (
   <div>
-    {title && generateTitle(title)}
-    <StyledCarousel>{generateChildren(cards, title)}</StyledCarousel>
+    {(() =>
+      cards.length > 0 ? (
```
val3ntin: could be `cards.length ?  A : B `

src/modules/Watchbar/index.js
```diff
@@ -80,8 +80,13 @@ const generateChildren = (cards, title) => {
 
 const WatchBar = ({ cards, title }) => (
   <div>
```
val3ntin: do we need ever to render this `div` here in case no cards are rendered ?

src/modules/Watchbar/index.js
arturkin: Done

src/modules/Watchbar/index.js
arturkin: Done 👍

PR #102, opened 2019-04-08
EWL-499 Implement visual regression test via Applitools

davidseow:
Screen Shot 2019-04-08 at 11 18 58

The additional step in the pipeline will kick off visual regression test using Applitools. The implementation latches on the existing workflow of creating Storybook stories by doing visual snapshots of our components and then comparing them with a base stored in Applitools.

👉👉**Once this is merged in, our workflow will then need to include reviewing and updating visual snapshots in Applitools.**👈👈

Screen Shot 2019-04-08 at 11 34 16

Screen Shot 2019-04-09 at 10 54 12


val3ntin:
👍 
@davidseow could you please provide example for responsive UI tests as well ? 
the breakpoints can be found into the `./src/breakpoints` file

davidseow:
@val3ntin sure, this can be done by passing browsers to run against in `applitools.config.js`. For example:
```JavaScript
module.exports = {
  browser: [
    {
      width: 1280,
      height: 1024,
      name: 'chrome',
    },
    {
      width: 480,
      height: 320,
      name: 'chrome',
    },
  ],
};

```
Details on this available via  [link](https://github.com/applitools/eyes-storybook#configuring-the-browser). 

Note, this will double up all the test since this is a direct integration via Storybook. If we need to capture multiple breakpoints for a specific component, we'll need to write those as a scenario in cucumberJs

.storybook/config.js
```diff
@@ -14,7 +14,9 @@ function loadStories() {
   req.keys().forEach(filename => req(filename));
 }
 
+const { STORYBOOK_OVERRIDE_BG } = process.env;
 const StyledWrapper = styled.div`
+  background-color: ${() => (STORYBOOK_OVERRIDE_BG ? '#92b900' : 'inherit')};
```
davidseow: Applitools eyes.storybook accesses component directly within its iframe bypassing addon-background color, setting background to a non-common colour so components will be visible

.circleci/config.yml
```diff
@@ -34,6 +34,19 @@ jobs:
...
       - run: npm run lint
+  test_visual:
+    <<: *defaults
+    steps:
+      - checkout
+      - *restore_cache
+      - run: 'export APPLITOOLS_BATCH_ID=`echo $CIRCLE_SHA1`'
+      - run:
+          command: |
+            if [[ -n $ENABLE_VISUAL_TEST ]]; then
```
davidseow: disable test by default so we could get this in before trial period starts. we'll need to add this env variable to CI to enable it

.circleci/config.yml
```diff
@@ -46,3 +59,18 @@ workflows:
       - lint:
           requires:
             - dependencies
+      - run_visual_test:
```
davidseow: As per discussion, visual test can be triggered manually on master branch as long as lint and unit test is green

PR #104, opened 2019-04-11
Ewl 471 create adobe component

ra-phael:


src/components/Adobe/index.js
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+class Adobe extends React.Component {
+  constructor(props) {
+    super(props);
+    const { src } = this.props;
+    this.adobeScript = document.createElement('script');
```
val3ntin: within all code below you're referring to `document` into the constructor whilst you won't have a document during the SSR.

val3ntin: if we consider analytics as an important part of the application, i assume we want it to be rendered during the SSR as well.

thus maybe try re-writing component to return the `<script>` element so we could use it directly put it into the head ?

maybe @leozero has some thoughts on it as well.

src/components/Adobe/index.js
```diff
@@ -0,0 +1,46 @@
...
+    this.adobeScript.async = false;
+
+    this.satelliteScript = document.createElement('script');
+    this.satelliteScript.id = 'satellite-script';
+    this.satelliteScript.type = 'text/javascript';
+    this.satelliteScript.async = false;
+    this.satelliteScript.innerHTML = `if(typeof(_satellite) != 'undefined'){_satellite.pageBottom();}`;
+
+    document.head.appendChild(this.adobeScript);
+    document.head.appendChild(this.satelliteScript);
```
val3ntin: satellite script should be added to the page's body, not head.

src/components/Adobe/index.test.js
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import { shallow, mount } from 'enzyme';
+import { Adobe } from '../..';
+
+describe('<Adobe />', () => {
+  it('renders <Adobe />', () => {
+    const component = shallow(<Adobe src="srcUrl" />);
+    expect(component).toMatchSnapshot();
+    component.unmount();
```
val3ntin: no need to do this here i think as you don't referring to `componentWillUnmount` lifecycle

src/components/Adobe/index.js
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+class Adobe extends React.Component {
+  constructor(props) {
+    super(props);
+    const { src } = this.props;
+    this.adobeScript = document.createElement('script');
+    this.adobeScript.id = 'adobe-script';
+    // this.adobeScript.type = 'text/javascript';
```
val3ntin: remove this if not used ;)

src/components/Adobe/index.js
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+class Adobe extends React.Component {
+  constructor(props) {
+    super(props);
+    const { src } = this.props;
```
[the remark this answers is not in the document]
ra-phael: Agreed 👍

src/components/Adobe/index.test.js
ra-phael: Are you refering to component.unmount(); ? I added it because otherwise the component is still mounted and can conflict with the subsequent tests: while testing for the src url in my 3rd tests I was still getting the url set in my first test...

src/components/Adobe/index.js
ra-phael: Thanks 🙈

PR #106, opened 2019-04-15
EWL-505 Added cypress

davidseow:
Screen Shot 2019-04-15 at 09 00 06


.circleci/config.yml
```diff
@@ -21,7 +21,8 @@ jobs:
       - save_cache:
           paths:
             - node_modules
-          key: v1-dependencies-{{ checksum "package-lock.json" }}
+            - ~/.cache/Cypress
+          key: v2-dependencies-{{ checksum "package-lock.json" }}
```
davidseow: had to bump up the version since there is no updates to package-lock.json and I needed it to cache cypress folder

.circleci/config.yml
```diff
@@ -34,12 +35,28 @@ jobs:
...
       - *restore_cache
       - run: npm run lint
+  test_cypress:
+    <<: *defaults
+    steps:
+      - checkout
+      - *restore_cache
+      - run: npm run build-storybook
+      - run:
+          command: npm run start-static-storybook
```
davidseow: boots up static storybook

cypress/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
```
davidseow: enable autocompletion in vscode

cypress.json
```diff
@@ -0,0 +1,5 @@
+{
+  "baseUrl": "http://localhost:5000",
+  "video": false,
+  "trashAssetsBeforeRuns": true
```
leozero: By default, the value is true so I think we don't need this line ;-)
(see https://docs.cypress.io/guides/references/configuration.html#Screenshots)

PR #107, opened 2019-04-15
Ewl 476 create quantcast component

ra-phael:


ra-phael:
I'm not sure to understand why snapshots from other components that I didn't touch often appeared as modified. 🤔
After updating the snapshots and with all tests passing on my end, they still seem to fail in CircleCI.

[Jest-styled-components](https://github.com/styled-components/jest-styled-components) seems to provide more helpful indications on why a snapshot is different (beyond just the classNames hashes differences).

PR #116, opened 2019-04-26
Add npm jobs/workflow

davidgarry:
https://circleci.com/blog/publishing-npm-packages-using-circleci-2-0/


.circleci/config.yml
```diff
@@ -61,6 +61,30 @@ jobs:
...
+      - run: npm publish
+      - run: git push
+  npm_major:
+    <<: *defaults
+    steps:
+      - checkout
+      - *restore_cache
+      - run: npm version major
+      - run: npm publish
+      - run: git push
```
val3ntin: do we need `git push` for each described job ? Since we're running this on a master branch already ?

val3ntin: ah,... right... need to keep it within a branch. 👍

davidgarry: @val3ntin We want to push the new commit created by the npm version xxx command.

PR #125, opened 2019-05-06
EWL-539-Quickpoll-component

ra-phael:
https://eurosportdigital.atlassian.net/browse/EWL-539


src/components/Quickpoll/index.js
```diff
@@ -0,0 +1,149 @@
...
+import * as colors from '../../colors';
+import { fontFamilies } from '../../typography';
+import { Button } from '../..';
+
+const StyledContainer = styled.div`
+  border-radius: 4px;
+  background-color: ${colors.flawlessMahogany};
+  position: absolute;
+  width: 100%;
+  height: 392px;
```
val3ntin: i'm not sure this works well as:
1. this is ablolutely-positioned div with fixed heigh.
the height of the DIV won't be re-calculated 
and once you have more answers in quickpoll - it's broken.

Screen Shot 2019-05-06 at 2 11 30 PM

src/components/Quickpoll/index.js
```diff
@@ -0,0 +1,149 @@
...
+export const StyledTitle = styled.div`
+  position: relative;
+  font-size: 20px;
+  font-family: ${fontFamilies.alphaHeadline};
+  margin: 26px 19px;
+  z-index: 2;
+`;
+
+export const StyledChoices = styled.div`
+  position: absolute;
```
val3ntin: just curious why we need all of these absolutely-positioned ?
the parent container won't grow up / down in case you have less/more items...

val3ntin: if you need z-index to work - you may have a related position or even an `opacity / transition` property for your element...
https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context#The_stacking_context

src/components/Quickpoll/index.js
```diff
@@ -0,0 +1,149 @@
...
+  ${props =>
+    props.showResults &&
+    css`
+      padding: 2px 0;
+      border: 0px;
+      border-radius: 0;
+      background-color: ${colors.actionOneDarkMinus1};
+      height: 48px;
+      :hover {
+        color: inherit;
```
val3ntin: and I think override default `hover: cursor` as well

src/components/Quickpoll/index.js
```diff
@@ -0,0 +1,149 @@
...
+  };
+
+  render() {
+    const { title, choices, showResults } = this.props;
+    return (
+      <StyledContainer>
+        <StyledBackground />
+        <StyledTitle>{title}</StyledTitle>
+        <StyledChoices>
+          {choices.map(element => (
```
val3ntin: i think for consistency it can be renamed to `choice`

src/components/Quickpoll/index.js
```diff
@@ -0,0 +1,149 @@
...
+  handleClick = id => {
+    const { showResults, onChoiceClick } = this.props;
+    if (showResults !== true) {
+      onChoiceClick(id);
+    }
+  };
+
+  render() {
+    const { title, choices, showResults } = this.props;
+    return (
```
val3ntin: are we able to render either `choices` or `results` here ?

src/components/Quickpoll/index.js
ra-phael: I tried to avoid: {choice.choice}
Maybe then I could rename the property to have {choice.displayText} or something like that.

src/components/Quickpoll/index.js
ra-phael: Yes, based on the boolean showResults prop.
As far as data, it's the same `choices`, only with some additional voting data.

src/components/Quickpoll/index.js
val3ntin: oops, didn't notice `choice.choice` :)

src/components/Quickpoll/index.js
val3ntin: ok, i see. thanks

src/components/Quickpoll/index.js
ra-phael: fixed

src/components/Quickpoll/index.js
```diff
@@ -0,0 +1,148 @@
...
+  padding-top: 10px;
+`;
+
+const StyledBackground = styled.div`
+  position: absolute;
+  height: 100%;
+  width: 100%;
+  border-radius: 4px;
+  background-color: ${colors.coreNeutral9};
+  clip-path: polygon(0 11%, 100% 0%, 100% 100%, 0% 100%);
```
val3ntin: honestly i'm not sure this diagonal will match the design diagonal angle in case it's integrated for different resolutions / width...

src/components/Quickpoll/Quickpoll.test.js
```diff
@@ -0,0 +1,81 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import Quickpoll, { StyledChoices, StyledChoiceText, StyledTitle, StyledResultPercentage } from '.';
+import { choices, choicesWithResults } from './index.stories';
```
davidgarry: It could be better to create a mock file instead of loading the stories.

src/components/Quickpoll/index.stories.js
```diff
@@ -0,0 +1,40 @@
...
+
+export const choicesWithResults = [
+  { id: 123451, displayText: 'Nadal', result: '50%' },
+  { id: 123452, displayText: 'Djokovic', result: '47%' },
+  { id: 123453, displayText: 'Zverev', result: '34%' },
+  { id: 123454, displayText: 'Federer', result: '16%' },
+  { id: 123455, displayText: 'Other', result: '1%' },
+];
+
+const quickPoll = storiesOf('Components|QuickPoll', module).addDecorator(withInfo);
```
davidgarry: Please remove `.addDecorator(withInfo)`. It's now handled globally ;)

src/elements/Spinner/index.js
```diff
@@ -0,0 +1,70 @@
...
+  }
+`;
+
+const StyledWrapper = styled.div`
+  margin: 10px;
+  width: ${props => props.width};
+  height: ${props => props.width};
+`;
+
+const Spinner = ({ color, width, className }) => (
```
ra-phael: className seems necessary here if we want to override some css with Emotion (https://github.com/emotion-js/emotion/issues/257)

PR #127, opened 2019-05-07
EWL-498: Advanced Ad component

arturkin:
https://eurosportdigital.atlassian.net/browse/EWL-498
HOC for `AdPlacement` which automatically refreshes Ads when user is on the page.
_Ported from twilight code (`adManager.js/visibilityHelper.js`)_

Rules:
* check for element is in view(not scrolled up/down)
* checks for element is not overlapped by other element
* listens to scroll/resize to restore timeout
* listens to page api to stop refresh on tab change


src/components/Advertisement/AdReload/index.js
```diff
@@ -0,0 +1,125 @@
...
+      });
+
+      document.addEventListener('visibilitychange', this.handleVisibilityChange);
+    }
+
+    refreshAd() {
+      const { current } = this.adRef;
+      const { refreshHandler } = this.props;
+
+      refreshHandler(current.id);
```
val3ntin: add test for this as well.
expect(mockHandler).toBeCalledWith(elementId)

src/components/Advertisement/AdReload/visibilityHelpers.js
```diff
@@ -0,0 +1,56 @@
...
+        el = el.parentElement || el.parentNode;
+      } while (el !== null && el.nodeType === 1);
+      return null;
+    };
+  }
+};
+
+closestPolyfill();
+
+export const isElement1ParentOfElement2 = (parent, child) => {
```
val3ntin: don't export if using inside this module only

src/components/Advertisement/index.stories.js
```diff
@@ -3,11 +3,18 @@ import { storiesOf } from '@storybook/react';
...
+import { HP_MAIN } from './AdManager/page-types';
+import { MPU } from './AdManager/ad-types';
+
+const AdWithReload = withReload(AdPlacement);
 
 const indexStories = storiesOf('Components|Advertisement', module).addParameters({
   backgrounds: { disable: true },
 });
 
+AdManager.defaultSlotsConfig.desktop[HP_MAIN][MPU].count = 7;
```
val3ntin: as you changed default slots config to 7 - then please remove the `extra slots` .... as I think all of 7 slots going to be rendered...
 - `extra desktop ad - out of homepage config`

src/components/Advertisement/AdReload/index.js
```diff
@@ -0,0 +1,125 @@
...
+import { isOnViewPort, isTopMost } from './visibilityHelpers';
+
+function cleanQueryString() {
+  if (window.location.search.indexOf('?_ga') > -1) {
+    window.history.replaceState(null, null, window.location.href.replace(window.location.search, ''));
+  }
+}
+
+const withReload = Component =>
+  class WithReload extends React.PureComponent {
```
val3ntin: would be useful to have displayname
e.g. :
https://github.com/EurosportDigital/web-toolkit/blob/9ac9db15413c1a5a791ad6bb8b0bda7429ddce02/src/hocs/withMatchMedia.js#L11

src/components/Advertisement/AdReload/index.js
arturkin: done

src/components/Advertisement/AdReload/visibilityHelpers.js
arturkin: done

src/components/Advertisement/index.stories.js
arturkin: done; moved reloaded ads upper, and set slots to 6. So extra slot will be empty.

src/components/Advertisement/AdReload/index.js
arturkin: done

src/index.js
```diff
@@ -83,4 +84,5 @@ export {
   AdInit,
   AdManager,
   AdPlacement,
+  withReload,
```
davidgarry: As withReload is specific to ads, maybe it could be better to add the term "ad" in the export name ?

arturkin: I did this component not specific to Ads, it may be used to run any callback on any react element. That's why it doesn't have `ad` in its name.

PR #133, opened 2019-05-10
Ewl 535 create player cards

ra-phael:
https://eurosportdigital.atlassian.net/browse/EWL-535


src/modules/PlayerCard/index.js
```diff
@@ -0,0 +1,99 @@
...
+  ul {
+    margin-bottom: 45px;
+  }
+`;
+
+const PlayerCard = ({ playerInfo, backgroundImageUrl, previousMatches }) => (
+  <StyledContainer>
+    <StyledBackground backgroundImageUrl={backgroundImageUrl} />
+    <StyledPlayerInfos player={playerInfo} />
+    <StyledPreviousMatches>PREVIOUS MATCHES</StyledPreviousMatches>
```
davidgarry: Display it only if there are some previous matches

src/modules/PlayerCard/index.js
```diff
@@ -0,0 +1,99 @@
...
+    <StyledMatches>
+      <ViewMore showLessText="View Less Matches" showMoreText="View More Matches">
```
davidgarry: The ViewMore button is broken on the storybook, it displays 4 items, and when I click on the button, it does not display the last item.

src/modules/PlayerCard/index.js
```diff
@@ -0,0 +1,99 @@
...
+);
+
+const matchType = PropTypes.shape({
+  id: PropTypes.number.isRequired,
+  matchUrl: PropTypes.string.isRequired,
+  data: PropTypes.shape({
+    bottomTeam: teamDataType,
+    topTeam: teamDataType,
+  }).isRequired,
+  displayLeftCircle: PropTypes.oneOf(['won', 'lost', false]),
```
davidgarry: This property is not used as your code use `match.hasWon`
hasWon property is missing

src/modules/PlayerCard/index.js
```diff
@@ -0,0 +1,99 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'react-emotion';
+import * as colors from '../../colors';
+import { fontFamilies } from '../../typography';
+import PlayerInfos, { playerType } from '../../components/PlayerInfos';
+import ScoreBlock from '../../components/ScoreBlock';
+import { teamDataType } from '../../components/ScoreBlock/ScoreBlock';
```
davidgarry: It should be better to import it from `../../components/ScoreBlock`

src/modules/PlayerCard/index.js
davidgarry: The View More button must be aligned with the player names and not centered 👍 
image

src/modules/PlayerCard/index.js
ra-phael: Ok, I'll export teamDataType from ScoreBlock's index.js

src/modules/PlayerCard/index.js
ra-phael: Thanks, I developed for mobile first and didn't notice this difference on desktop!

PR #135, opened 2019-05-10
EWL 545#2 Detecting Window Resize on ViewMore Component to update its height

saad-kh:
The ViewMore Component should have cared about the resizing of the Window because a new Window`size can make its displayed height less accurate.
This PR fix that by decorating it with an external [package ](https://github.com/maslianok/react-resize-detector) HOC that rerenders the ViewMore component on window resize.


arturkin:
Do we really need a new dependency instead of 
```
window.addEventListener(
   'resize', 
   throttle(()=> this.setState({windowWidth: window.innerWidth})), 500);
```

saad-kh:
@arturkin  
You're right and that was the initial plan. However without the dependency we'd need to add and remove the event Listener (cleanup code), and make the setState call that doesn't have an explicit meaning but is here only for the rerendering side effect. The dependency let us reason about it with a more straightforward React paradigm.

arturkin:
43Kb gzipped dependency with it's own deps for window.onresize it too much even in react word 😄 
http://maslianok.github.io/react-resize-detector/static/js/main.ef81d4ad.js

saad-kh:
Wouw that's too much indeed. You're right I'll revert to a simple addEventListener.

saad-kh:
@Xb0X @arturkin Updated

src/modules/ViewMore/index.js
```diff
@@ -40,8 +40,15 @@ export default class ViewMore extends React.Component {
     // It is better to set the default value of expanded after the initial mount.
     // Otherwise the first render will not be accurate as the listRef might not be set yet.
     this.setState({ expanded: false });
+    window.addEventListener('resize', this.handleWindowResize);
```
arturkin: Please add throttle/debounce

src/modules/ViewMore/index.js
```diff
@@ -40,8 +40,15 @@ export default class ViewMore extends React.Component {
...
   }
 
+  componentWillUnmount() {
+    window.removeEventListener('resize', this.handleWindowResize);
+  }
+
+  handleWindowResize = () => this.setState({});
```
arturkin: You will loose `expanded` state here?

saad-kh: No it'll preserve the current state.

saad-kh:
@arturkin Throttle added

PR #143, opened 2019-05-14
Ewl 534 match hero block

ra-phael:
https://eurosportdigital.atlassian.net/browse/EWL-534

Ended up splitting into 2 different components as they require different sets of props.


src/modules/MatchHero/index.js
```diff
@@ -0,0 +1,149 @@
...
+`;
+
+const MatchHero = ({ title, date, hour, labels, cTAText, cTALink, marketingMessage }) => (
+  <StyledWrapper>
+    <Labels labels={labels} />
+    <StyledContent>
+      <StyledBasicInfo>
+        <H1>{title}</H1>
+        <StyledTime>
+          {date} | Starting from {hour}
```
arturkin: Why not to use label here?

src/modules/MatchHero/index.js
```diff
@@ -0,0 +1,149 @@
...
+  hour: '',
+  marketingMessage: '',
+};
+
+MatchHero.propTypes = {
+  title: PropTypes.string.isRequired,
+  date: PropTypes.string,
+  hour: PropTypes.string,
+  labels: labelsType.isRequired,
+  cTAText: PropTypes.string.isRequired,
```
arturkin: Caps is typo or it's intentional?

src/modules/MatchHero/index.js
```diff
@@ -0,0 +1,149 @@
...
+
+export const MatchHeroWithScore = ({ gameStatus, scoreData, labels, displayWatchButton, onWatchButtonClick }) => (
+  <StyledWrapper>
+    <Labels labels={labels} />
+    <StyledScoreWrapper>
+      <SetsScore data={scoreData} />
+    </StyledScoreWrapper>
+    {displayWatchButton && (
+      <PlayIconLink bgColorIcon={colors.utahCrimson} bgColorText={colors.venetianRed} onClick={onWatchButtonClick}>
+        {gameStatus === 'after' ? 'WATCH REPLAY' : 'WATCH LIVE'}
```
arturkin: Why not to use labels passed from props?

src/modules/MatchHero/index.js
ra-phael: The Labels component?

src/modules/MatchHero/index.js
ra-phael: Intentional as it's an acronym (Call To Action) but I don't have strong opinions about that. :)

src/modules/MatchHero/index.js
ra-phael: Good idea, that'll remove some logic from this component.

src/modules/MatchHero/index.js
arturkin: No I meant string should not be hardcoded

src/modules/MatchHero/index.js
```diff
@@ -0,0 +1,149 @@
...
+  min-width: 282px;
+  max-width: 368px;
+  ${breakpoints.medium(css`
+    margin-bottom: 43px;
+  `)};
+`;
+
+const MatchHero = ({ title, date, hour, labels, cTAText, cTALink, marketingMessage }) => (
+  <StyledWrapper>
+    <Labels labels={labels} />
```
arturkin: Please add picture component to display hero background-image!

src/modules/MatchHero/index.js
val3ntin: separator `|` should be styled here...

src/modules/MatchHero/index.js
ra-phael: After discussing with David Garry we thought it'd be better if the zone is in charge of injecting the player within itself and therefore also handle the display of a background image in the right place.

src/modules/MatchHero/index.js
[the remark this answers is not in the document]
ra-phael: I removed gameStatus as it wasn't used at all

src/modules/MatchHero/index.js
ra-phael: done

src/modules/MatchHero/index.test.js
```diff
@@ -0,0 +1,138 @@
...
+        scoreData={scoreDataDuring}
+        displayWatchButton
+        onWatchButtonClick={() => null}
+        watchButtonText="WATCH LIVE"
+      />
+    );
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it('should display the watch button', () => {
```
val3ntin: i think this test can be removed as watchButton display already into your snapshot from line 72

src/modules/MatchHero/index.test.js
```diff
@@ -0,0 +1,138 @@
...
+        scoreData={scoreDataDuring}
+        displayWatchButton
+        onWatchButtonClick={() => null}
+      />
+    );
+    const playButton = wrapper.find({ bgColorText: colors.venetianRed });
+    expect(playButton).toHaveLength(1);
+  });
+
+  it('should display a text on watch button passed as props', () => {
```
val3ntin: the same - tested as a part of your snapshot.

src/modules/MatchHero/index.test.js
```diff
@@ -0,0 +1,138 @@
...
+        marketingMessage="Watch Eurosport live"
+      />
+    );
+  });
+
+  it('should render as expected', () => {
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it('should render labels', () => {
```
val3ntin: remove it - it's tested as a part of the snapshot above ( line 29)

src/modules/MatchHero/index.test.js
```diff
@@ -0,0 +1,138 @@
...
+    expect(wrapper.find(Labels)).toHaveLength(1);
+  });
+
+  it('should render a CTA', () => {
```
val3ntin: tested as a part of the snapshot above ( line 29)

src/modules/MatchHero/index.test.js
```diff
@@ -0,0 +1,138 @@
...
+    expect(wrapper.find(StyledCTA)).toHaveLength(1);
+  });
+
+  it('should render a marketing message', () => {
```
val3ntin: tested as a part of the snapshot above ( line 29)

src/modules/MatchHero/index.test.js
```diff
@@ -0,0 +1,138 @@
...
+    expect(
+      wrapper
+        .find(StyledMarketingMessage)
+        .childAt(0)
+        .text()
+    ).toEqual('Watch Eurosport live');
+  });
+
+  it('should render Basic Info', () => {
```
val3ntin: tested as a part of the snapshot above ( line 29)

src/modules/MatchHero/index.test.js
```diff
@@ -0,0 +1,138 @@
...
+        onWatchButtonClick={() => null}
+        watchButtonText="WATCH REPLAY"
+      />
+    );
+    const playButton = wrapper.find({ bgColorText: colors.venetianRed });
+    expect(playButton).toHaveLength(1);
+    expect(playButton.prop('children')).toEqual('WATCH REPLAY');
+  });
+
+  it('should render SetScore', () => {
```
val3ntin: covered in your snapshot as well ?

arturkin:
Last thing I noticed - it's a bad text wrap for the tablet and mobile views:
image

PR #151, opened 2019-05-17
Ewl 581 all matches

ra-phael:
https://eurosportdigital.atlassian.net/browse/EWL-581


src/modules/AllMatches/index.js
```diff
@@ -0,0 +1,216 @@
+import React from 'react';
+import styled, { css } from 'react-emotion';
+import PropTypes from 'prop-types';
+import * as breakpoints from '../../breakpoints';
```
leozero: Import only colors and breakpoints needed

src/modules/ViewMore/index.js
```diff
@@ -7,7 +7,9 @@ import throttle from 'lodash/throttle';
 import { coreLightMinus1, coreLightBase, coreDarkPlus1 } from '../../colors';
 import { fontAlphaHeadlineBold } from '../../typography';
 
-export const StyledButton = styled.span`
+export const StyledViewMoreButtonWrapper = styled('div')``;
```
leozero: You can change for styled.div instead of styled('div') (as below)

PR #155, opened 2019-05-17
Add i18n and example in DeactivatedComments

leozero:
https://eurosportdigital.atlassian.net/browse/EWL-583

Example in deactivatedComments module


src/modules/UserComments/DeactivatedComments.js
```diff
@@ -24,11 +26,15 @@ const StyledIcon = styled.img`
...
+    <div>{t('match_page.usercomments.deactivated_message')}</div>
   </StyledWrapper>
 );
 
-export default DeactivatedComments;
+DeactivatedComments.propTypes = {
+  t: PropTypes.func.isRequired,
+};
+
+export default withTranslation()(DeactivatedComments);
```
val3ntin: i think for tests you should be exporting plain component.
as jest snapshot it not informative now: you miss SVG  `StyledIcon `

src/modules/UserComments/index.test.js
```diff
@@ -30,6 +30,6 @@ describe('UserComments', () => {
     wrapper.setProps({
       livefyreConfig: { deactivatedCommentsConfig },
     });
-    expect(wrapper.find('DeactivatedComments')).toHaveLength(1);
+    expect(wrapper.find('DeactivatedComments')).toBeDefined();
```
val3ntin: will not work. as  `find` returns object and it's always defined.
test the plain component without i18N provider

src/modules/UserComments/__snapshots__/DeactivatedComments.test.js.snap
```diff
@@ -1,13 +1,9 @@
...
-<Styled(div)>
-  <Styled(img)
-    alt="lock icon"
-    src="lock-icon.svg"
-  />
-  <div>
-    Comments are deactivated for this publication
-  </div>
-</Styled(div)>
+<DeactivatedComments
```
val3ntin: snapshot it not informative now.
we should keep the usual component for test ( which should not be wrapped into the `withTranslation()` )

src/modules/UserComments/DeactivatedComments.js
val3ntin: just fixed this on my own

PR #158, opened 2019-05-18
Remove matchHero paddings, will be handled with Grid

davidgarry:


src/modules/MatchHero/index.js
```diff
@@ -10,23 +10,10 @@ import Button from '../../elements/Button';
...
-  `)};
-
-  ${breakpoints.medium(css`
-    padding: 30px 34px;
-  `)};
-
-  ${breakpoints.wide(css`
-    padding: 30px 150px;
-  `)};
+  padding: 30px 0;
```
david-sautet: Is it normal to keep vertical padding?
Is it not better to manage it in ETS?

davidgarry: Fixed + added support for className

PR #190, opened 2019-05-23
change proptypes for title to allow html node to style VS

Xb0X:


src/modules/MatchHero/index.js
```diff
@@ -120,7 +120,7 @@ MatchHero.defaultProps = {
 
 MatchHero.propTypes = {
   className: PropTypes.string,
-  title: PropTypes.string.isRequired,
+  title: PropTypes.node.isRequired,
```
arturkin: Maybe oneOf string/node ?

davidgarry: @arturkin In fact, PropTypes.node means all that can be rendered by react, it can be a react fragment, a string, a number...

davidgarry: Extract from the PropTypes doc : 
```
// Anything that can be rendered: numbers, strings, elements or an array
  // (or fragment) containing these types.
  // see https://reactjs.org/docs/rendering-elements.html for more info
  optionalNode: PropTypes.node,
```
https://github.com/facebook/prop-types

arturkin: Right @davidgarry

PR #199, opened 2019-05-27
(EWL-475): Add Adobe video events

JulienKode:
# Add Adobe video events

with @david-sautet 

## Description

We made a little component `HeartbeatInit` just to make sure the Heartbeat script is loaded
When this is loaded we use an `withVideoAnalytics` Hoc with JWPlayer to add analytics 
In addition, we've also added a class to manage Adobe Heartbeat events (`AdobeHeartbeatAnalytics`)

## Tests 

We tested `AdobeHeartbeatAnalytics` business logic
We tested `withVideoAnalytics` to make sure he calls the right function


src/components/Adobe/index.stories.js
```diff
@@ -1,8 +1,48 @@
...
+
+const innerHTML = `
+if(typeof(_satellite) != 'undefined'){
+  _satellite.pageBottom();
+}
+`;
+
+indexStories
+  .add('Adobe', () => <Adobe src={text('src', 'url.toscript.js')} isServerSide={false} />)
+  .add('with JWPPlayer', () => (
```
davidgarry: Typo : JWPlayer

src/lib/AdobeHeartbeatAnalytics/AdobeHeartbeatAnalytics.js
```diff
@@ -0,0 +1,134 @@
...
+
+  calculateCurrentPlaybackTime = (initialPlaybackTime, stopwatch) => () => initialPlaybackTime + stopwatch.ms / 1000;
+
+  onAdBreakStart = (name = 'preroll', position = 1) => {
+    const { MediaHeartbeat, mediaHeartbeat } = this;
+
+    const startTime = 0;
+    const adBreakObject = MediaHeartbeat.createAdBreakObject(name, position, startTime);
+
+    mediaHeartbeat.trackEvent(MediaHeartbeat.Event.AdBreakStart, adBreakObject);
```
davidgarry: `mediaHeartbeat && ` missing

src/lib/AdobeHeartbeatAnalytics/AdobeHeartbeatAnalytics.js
```diff
@@ -0,0 +1,134 @@
+import Stopwatch from 'timer-stopwatch';
```
arturkin: This module seems to be working only with SSR as it requires node:
https://github.com/MickCrozier/timer-stopwatch/blob/master/lib/Stopwatch.js#L6
Also, it requires old lodash module.. 

Maybe we can replace it with setTimeout + Date.now() ?
Maybe we can replace it with just

src/lib/AdobeHeartbeatAnalytics/AdobeHeartbeatAnalytics.js
```diff
@@ -0,0 +1,134 @@
...
+
+export default class AdobeHeartbeatAnalytics {
+  constructor(heartbeatConfig, videoMetadata, programStartDateTime) {
+    const { ADB } = window;
```
arturkin: This will crash with SSR

src/lib/AdobeHeartbeatAnalytics/AdobeHeartbeatAnalytics.js
JulienKode: Thanks 😊

src/components/Adobe/HeartbeatInit/index.js
```diff
@@ -0,0 +1,35 @@
+import { Component } from 'react';
+import PropTypes from 'prop-types';
+
+class HeartbeatInit extends Component {
```
arturkin: Why do you need this component for? Seems that you can register callback in a `AdobeHeartbeatAnalytics.js` constructor, and this component will be redundant

src/lib/AdobeHeartbeatAnalytics/AdobeHeartbeatAnalytics.js
```diff
@@ -0,0 +1,134 @@
...
+    const { MediaHeartbeat, MediaHeartbeatConfig, MediaHeartbeatDelegate } = ADB.va;
+
+    const mediaConfig = this.generateMediaConfig(MediaHeartbeatConfig, heartbeatConfig, videoMetadata);
+    const appMeasurement = window.s_c_il[1];
```
arturkin: This will crash with SSR

src/lib/AdobeHeartbeatAnalytics/AdobeHeartbeatAnalytics.js
JulienKode: This will be instantiate in the client side

src/components/Adobe/HeartbeatInit/index.js
JulienKode: It is just a tool to know if Adobe Heartbeat SDK was loaded, 
Then we can instantiate `AdobeHeartbeatAnalytics`

src/lib/AdobeHeartbeatAnalytics/AdobeHeartbeatAnalytics.js
JulienKode: Ok w've try based on the Eurosport-react-app from the London team, and It seems to work properly

arturkin: I did clean test with webpack’ed stopwatch module, and it builds up to 36 kb minified code, because of node EventEmitter.. We can remove it later thought.

JulienKode: @arturkin Thanks for your analysis

PR #200, opened 2019-05-27
EWL-667 PlayerInfos visual adjustments

arturkin:
* Update design according to Kristian feedbacks
* Add support for 2 lines for player name
* Add external trigger to ViewMore (to sync 2 or more ViewMore expand)
* Export withMatchMedia HOC for external usage


src/components/PlayerInfos/index.js
```diff
@@ -171,9 +181,7 @@ export const PlayerInfos = ({ player, heightText, weightText, ageText, rankingTe
               player.firstName.length > player.lastName.length ? player.firstName.length : player.lastName.length
             }
           >
-            {player.firstName}
-            <br />
```
Xb0X: I think we need to keep this <BR/> ;) we discussed of it with Abdel and Maxime... Its because if one of the card is with the player name on one line and not the other the cards won't have the same height. And with the font size changing according to the name lenght, we handle a max of case to avoid 3 lines vs 2 ;)

arturkin: Thx for noticing this, will try to fix it

arturkin: Done. Now component can grow to height of the parent block

PR #220, opened 2019-06-11
reposition shop icon

ra-phael:


ra-phael:
Yes I wasn't sure about an hover effect since the idea is also to replicate the design of the watch pages which doesn't have one :)

PR #227, opened 2019-06-28
Ewl 807 toggle switch component

ra-phael:
https://eurosportdigital.atlassian.net/browse/EWL-807

### DIFF
--- a/src/elements/Toggle/index.js
+++ b/src/elements/Toggle/index.js
@@ -1,0 +1,89 @@
+import React from 'react';
+// eslint-disable-next-line no-unused-vars
+import PropTypes from 'prop-types';
+import styled from 'react-emotion';
+import { coreLightMinus1, azureRadiance } from '../../colors';
+import { fontInterUi } from '../../typography';
+
+const StyledToggleWrapper = styled.div`
+  text-align: center;
+  border-bottom: 1px solid rgba($white, 0.2);
+  max-width: 964px;
+  margin: 0 auto;
+  padding-bottom: 20px;
+  & > * {
+    vertical-align: middle;
+    display: inline-block;
+  }
+`;
+
+const StyledLabel = styled.span`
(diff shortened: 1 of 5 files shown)


src/elements/Toggle/index.js
```diff
@@ -0,0 +1,93 @@
...
+    background-color: ${azureRadiance};
+    transition: 0.4s;
+    border-radius: 50%;
+    ${props => (props.isSetToLeft ? '' : 'transform: translateX(18px);')}
+  }
+`;
+
+const Toggle = ({ toggleCallback, isSetToLeft, leftLabel, rightLabel }) => (
+  <StyledToggleWrapper>
+    <StyledLabel isActive={isSetToLeft} style={{ textAlign: 'right' }}>
```
val3ntin: just curious, but why don't move the inline `style={....}` directly to the `StyledLabel` component .e.g. by adding `isLeft` boolean prop ?

`<StyledLabel .... isLeft />`

ra-phael: yes, I considered doing that as well... I think the amount of code and complexity added is about the same. Unless we 100% forbid inline style, here it seems to be a good use case for it.

ra-phael:
Thanks for your feedback @val3ntin ! I just got words that actually we don't want any transition.

src/elements/Toggle/index.js
```diff
@@ -0,0 +1,87 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'react-emotion';
+import { coreLightMinus1, azureRadiance } from '../../colors';
```
JulienKode: Why we do not use a theme for the colors ? 

inside the style we can do something like:
```
const StyledStuff = styled.div`
    background-color: ${( { theme } ) => theme.myFavoriteColor.pink};
 `;
```

We will never want a white theme ?

PR #229, opened 2019-07-02
Ewl 816 create program detail component

ra-phael:
https://eurosportdigital.atlassian.net/browse/EWL-816


src/elements/ProgramDetails/index.js
```diff
@@ -0,0 +1,62 @@
...
+  color: ${suitGray};
+  * :nth-child(2) {
+    margin-left: 10px;
+    padding-left: 10px;
+
+    padding: 8px;
+  }
+`;
+const StyledIconWrapper = styled.div`
+  padding: ${props => `${props.iconPadding}px`};
```
val3ntin: sorry, i don't see a way of using this in a future.
once you develop a mobile-first app it's ok to use StyledComponents.
why we do need to set up a specific font size ?

for instance, then after component's usage you would need to subscribe for screen resize / mediachange and set the different font ?

I think once we have styledComponents here - we should follow the same approach.

However, if you want to have e.g. baseFontSize - then i'd advise using  baseFont and provide other values in e.g. `em`s

src/elements/ProgramDetails/index.test.js
```diff
@@ -0,0 +1,21 @@
...
+import ProgramDetails from '.';
+import ChannelIcon from '../ChannelIcon';
+
+describe.only('PlayIcon', () => {
+  it('should render as expected', () => {
+    const wrapper = shallow(<ProgramDetails textDetail="Text details" />);
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it('should render the default channel icon', () => {
```
val3ntin: rudimental test - it's already in your snapshot i think.

ra-phael:
Thanks for the feedback, originally I added Channel Icon which itself requires a height in px and reproduced the same approach for css properties. Even though accessing props in css with emotion is easy, the opposite is not... indeed we would need an HOC (or hook) to get the viewport size as props as you said.
So I made some changes to improve responsiveness, matching as much as possible legacy and removed the css-related props.
Let me know what you think @val3ntin

src/elements/ProgramDetails/index.js
```diff
@@ -0,0 +1,69 @@
...
+export const StyledText = styled.div`
+  padding: 10px;
+
+  ${medium(css`
+    padding: 15px;
+  `)};
+`;
+
+const ProgramDetails = ({ callsign, textDetail }) => (
+  <StyledDetails>
```
arturkin: Maybe it's better to add `className` prop here to have the ability to style this component with emotion

ra-phael: Added! :)

PR #232, opened 2019-07-03
Ewl 815 create livevideohero module

ra-phael:
https://eurosportdigital.atlassian.net/browse/EWL-815

I tried to reproduce the design of the legacy hero as much as possible, one slight change is a merge between head-player__details and head-player__description, also we now use a different font and labels.


src/modules/LiveVideoHero/index.js
```diff
@@ -0,0 +1,153 @@
...
+    position: absolute;
+    top: 0;
+    bottom: 0;
+    left: 0;
+    right: 0;
+  }
+
+  &::before {
+    opacity: .5;
+    background: radial-gradient(circle,rgba(14,20,61,0) 0,#0e143d 100%);
```
val3ntin: replace hardcode with predefined `blueZodiac` color

src/modules/LiveVideoHero/index.js
```diff
@@ -0,0 +1,153 @@
...
+    z-index: 1;
+  }
+
+  &::after {
+    opacity: .8;
+    background: linear-gradient(180deg,rgba(20,27,77,.03),#060825);
```
val3ntin: `brandPlus2` color name instead of hardcoded `#060825 `

src/modules/LiveVideoHero/index.js
```diff
@@ -0,0 +1,153 @@
...
+`;
+
+const StyledDetails = styled.div`
+  position: relative;
+  margin: 0 auto;
+  z-index: 3;
+
+  width: 83%;
+
+  > :nth-last-child(3) {
```
val3ntin: this is not very clear selector.
once you have less/more childs into the styledDetails it won't work the same.
by design i see it does adds pads/margins to have a bit distance for the play button.
Screen Shot 2019-07-04 at 12 32 28 PM

maybe render play icon out of `StyledDetails` and put StyledDetails to have `padding-top: 20px` and 1st its child to have margin-bottom of 7px ?

otherwise, i'd put styledDetails a top padding 20px and put 7px margin for the labels component only.

src/modules/LiveVideoHero/index.js
```diff
@@ -0,0 +1,153 @@
...
+      ${breakpoints.wide(css`
+        padding-top: 270px;
+      `)}
+  `}
+`;
+
+const StyledH1 = styled(H1)`
+  color: ${coreLightMinus1};
+  overflow-wrap: break-word;
+  margin-top: 0.2em;
```
val3ntin: i didn't notice font-size being defined into the one of parent components.
it means the `em` will be calculated based on body font size which may be slightly different.
maybe setup a font size for a parent of StyledH1 (`StyledDetails`) so we are more specific with it ?

src/modules/LiveVideoHero/index.test.js
```diff
@@ -0,0 +1,22 @@
...
+    labels: beforeEventLabels,
+    title: 'Athletics: Marathon des Sables',
+    programCallSign: 'E2NO',
+    programDetails: '14:00 - 16:00',
+    isPlayerLoading: false,
+    videoPlayerMode: false,
+  });
+
+  it('should render as expected', () => {
+    const wrapper = shallow(<LiveVideoHero {...makeProps()} />);
```
JulienKode: Should add a test with a different `videoPlayerMode` and then having a snapshot or tu ensure that we have the right `PlayIcon`

src/modules/LiveVideoHero/index.js
```diff
@@ -0,0 +1,153 @@
...
+    width: 800px;
+  `)}
+  ${breakpoints.wide(css`
+    width: 965px;
+  `)}
+
+  ${props =>
+    props.videoPlayerMode
+      ? css`
+          padding-top: 0px;
```
JulienKode: you can write ` padding-top: 0;` 
You do not need `px` when it is `0`

src/modules/LiveVideoHero/index.js
```diff
@@ -0,0 +1,153 @@
...
+  &::before {
+    opacity: .5;
+    background: radial-gradient(circle,rgba(14,20,61,0) 0,#0e143d 100%);
+    z-index: 1;
+  }
+
+  &::after {
+    opacity: .8;
+    background: linear-gradient(180deg,rgba(20,27,77,.03),#060825);
+    z-index: 2;
```
JulienKode: is it mandatory for this UI to rely on `z-index` ?
(I'm just wondering)

src/modules/LiveVideoHero/index.js
ra-phael: I'm not convinced with moving the Play Icon out of StyledDetails. I think it makes sense to have it there with the other elements, they form a whole block together.
I had tried to override Labels' css with emotion but that it didn't work, it was breaking the component, even after adding className to it. 🤔

src/modules/LiveVideoHero/index.js
ra-phael: Which font-size should that be? I took that from legacy and notice that there is font-size: 100% on all the parent components until <body>.

src/modules/LiveVideoHero/index.test.js
ra-phael: We usually take one 1 snapshot per component. Though the logic is pretty simple, you're right, I'm going to add a test for videoPlayerMode.

JulienKode: @ra-phael perfect as it is a spec that we can break during the refactoring. It will be nice to have that

src/modules/LiveVideoHero/index.js
ra-phael: I had the exact the same thought when I first had a look at the legacy hero, so I tried removing the z-index and you end up with the title etc... behind the gradients. As I said I tried to stay true to the original live page but yes they may be a better implementation...

JulienKode: @ra-phael ok no problem 😊 makes sense

src/modules/LiveVideoHero/index.js
val3ntin: sorry, confused myself and you. H1 does have font-size defined in it. all is ok as we just saw.
thx

PR #236, opened 2019-07-08
(EWL-849): Make Jest great again 

JulienKode:
## Description

- [x] Corejs 3 was not needed for `Storybook` however `Jest` need it. So we've bring it back as dependency.
- [x] Two change was made at the same time with #232 and #235 so we've updated this snapshot


JulienKode:
https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#corejs-3

PR #238, opened 2019-07-09
EWL-825 Round links

arturkin:
Adds SimpleTabs component
https://eurosportdigital.atlassian.net/browse/EWL-825


src/modules/SimpleTabs/index.test.js
```diff
@@ -0,0 +1,100 @@
...
+
+    wrapper.setProps({ onItemSelected: spy });
+    getChildren()
+      .at(1)
+      .simulate('click', { preventDefault: () => {} });
+
+    expect(spy).toHaveBeenCalledWith('#2', 1);
+  });
+
+  it('disabled and current tabs do not trigger onItemSelected callback', () => {
```
JulienKode: Should we add a test for `disabled` Tab ?

src/modules/SimpleTabs/index.js
```diff
@@ -0,0 +1,136 @@
...
+import PropTypes from 'prop-types';
+import styled, { css, keyframes } from 'react-emotion';
+import { coreLightMinus1, coreNeutral4 } from '../../colors';
+import { fontAlphaHeadline } from '../../typography';
+import * as breakpoints from '../../breakpoints';
+import Carousel, { StyledSlide } from '../../components/Carousel';
+import Link from '../../elements/Link';
+
+const StyledCarousel = styled(Carousel)`
+  justify-content: left;
```
val3ntin: the `left` value is only into a draft.
maybe replace it by `flex-start` ?

https://developer.mozilla.org/en-US/docs/Web/CSS/justify-content

Screen Shot 2019-07-10 at 9 31 22 AM

src/modules/SimpleTabs/index.js
```diff
@@ -0,0 +1,136 @@
...
+      PropTypes.shape({
+        label: PropTypes.string.isRequired,
+        disabled: PropTypes.boolean,
+        href: PropTypes.string,
+      })
+    ).isRequired,
+    className: PropTypes.string,
+  };
+
+  handleClick(e, href, tabIndex) {
```
val3ntin: this method will not work on the first mount.
meaning when you have 1st tab selected by default - consumer of your Tab would need to trigger `click` on its own which is weird.

please add default tab select logic into `componentDidMount` as well so we always have a callbac being executed once tab is selected by default.

val3ntin: something which is done in existing tabs:
https://github.com/EurosportDigital/web-toolkit/blob/master/src/modules/Tabs/index.js#L91

src/modules/SimpleTabs/index.js
```diff
@@ -0,0 +1,136 @@
...
+  render() {
+    const { tabs, initialTabIndex, className } = this.props;
+    const { selectedTabIndex } = this.state;
+    const currentTab = selectedTabIndex || initialTabIndex || 0;
+
+    return (
+      <StyledCarousel withArrow={false} className={className}>
+        {tabs.map(({ label, disabled, href }, i) => {
+          if (currentTab === i) return <TabSelected key={href}>{label}</TabSelected>;
+          if (disabled === true) return <TabInactive key={href}>{label}</TabInactive>;
```
val3ntin: `disabled: PropTypes.boolean,`
replace by `if (disabled)`

;)

src/modules/SimpleTabs/index.js
```diff
@@ -0,0 +1,136 @@
...
+
+  render() {
+    const { tabs, initialTabIndex, className } = this.props;
+    const { selectedTabIndex } = this.state;
+    const currentTab = selectedTabIndex || initialTabIndex || 0;
+
+    return (
+      <StyledCarousel withArrow={false} className={className}>
+        {tabs.map(({ label, disabled, href }, i) => {
+          if (currentTab === i) return <TabSelected key={href}>{label}</TabSelected>;
```
val3ntin: the `href` is not marked as required into propTypes definition.
therefore, `href`s can be undefined so all rendered tabs do have the same key.

please switch to more unique: `${label}${href}` ?

src/modules/SimpleTabs/index.js
```diff
@@ -0,0 +1,136 @@
...
+  }
+`;
+
+export default class SimpleTabs extends React.Component {
+  state = {
+    selectedTabIndex: null,
+  };
+
+  static defaultProps = {
+    initialTabIndex: null,
```
val3ntin: why not 0 by default ? 
this would remove the rudimental check on line 120.

```
 const currentTab = selectedTabIndex || initialTabIndex || 0;
```

src/modules/SimpleTabs/index.js
arturkin: Done

src/modules/SimpleTabs/index.js
arturkin: Good point, added call for callback on a DidMount

src/modules/SimpleTabs/index.js
arturkin: 💩

src/modules/SimpleTabs/index.js
arturkin: Done

src/modules/SimpleTabs/index.js
arturkin: 💩

src/modules/SimpleTabs/index.js
```diff
@@ -0,0 +1,136 @@
...
+
+  ${breakpoints.medium(css`
+    padding-bottom: 11px;
+    margin-right: 32px;
+  `)}
+
+  &:after {
+    content: '|';
+    position: absolute;
+    right: -13px;
```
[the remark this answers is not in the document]
arturkin: It's absolute, no way to position without magic numbers :)

src/modules/SimpleTabs/index.js
```diff
@@ -0,0 +1,136 @@
...
+    color: rgba(255, 255, 255, 0.25);
```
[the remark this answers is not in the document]
arturkin: I wanted to use just font color and not `opacity`.. No way to convert hex to rgba with emotion.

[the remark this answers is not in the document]
arturkin: Replaced with non-rgba color, we have similar one!

src/modules/SimpleTabs/index.test.js
arturkin: Did you mean to split this test? Because it check for disabled and current tabs.

src/modules/SimpleTabs/index.js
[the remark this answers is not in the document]
JulienKode: Oh nice 😊

src/modules/SimpleTabs/index.test.js
JulienKode: So it's good 😊

PR #239, opened 2019-07-09
Update liveVideoHero to manage dynamic min-height

Xb0X:
Here is a fix to remove the min-height when in player mode.

### DIFF
--- a/src/modules/LiveVideoHero/index.js
+++ b/src/modules/LiveVideoHero/index.js
@@ -12,7 +12,7 @@ const StyledLiveVideoHeroWrapper = styled.div`
   position: relative;
 
   & ::before,
-  ::after {
+  & ::after {
     content: '';
     display: block;
     position: absolute;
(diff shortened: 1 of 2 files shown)


src/modules/LiveVideoHero/index.js
```diff
@@ -11,8 +11,9 @@ import ProgramDetails from '../../elements/ProgramDetails';
 const StyledLiveVideoHeroWrapper = styled.div`
   position: relative;
 
-  & ::before, ::after {
-    content: "";
+  & ::before,
```
arturkin: looks like you're missing correct after selector here:
`& ::before, & ::after`

Xb0X: thx ;) it's done !!

src/modules/LiveVideoHero/index.js
```diff
@@ -22,30 +23,33 @@ const StyledLiveVideoHeroWrapper = styled.div`
...
-    min-height: 408px;
-  `)}
-  ${breakpoints.large(css`
-    min-height: 450px;
-  `)}
-  ${breakpoints.wide(css`
-    min-height: 580px;
-  `)}
+  ${props =>
+    !props.videoPlayerMode &&
```
JulienKode: Should we add a test or snapshot to test when the `props.videoPlayerMode` is here or not ?

Xb0X: we have the snap ;)

PR #256, opened 2019-07-16
Ewl 854 add label infos to player cards

degouville:
https://eurosportdigital.atlassian.net/browse/EWL-854


src/modules/PlayerCard/index.js
```diff
@@ -148,6 +166,7 @@ PlayerCard.propTypes = {
   className: PropTypes.string,
   onViewMoreClick: PropTypes.func,
   isViewMoreOpen: PropTypes.bool,
+  labels: labelsType.isRequired,
```
davidgarry: I'm sorry, it was not clear in the ticket and in the designs, but the labels can be different for each match. So you can't send just one instance of labels for all the matches.

src/modules/PlayerCard/index.js
```diff
@@ -111,9 +121,17 @@ export const PlayerCard = ({
...
             forceToggle={isViewMoreOpen}
           >
-            {previousMatches.map((match, index) => (
+            {previousMatches.map(match => (
               <StyledScoreBlock key={match.id}>
-                {index > 0 && <StyledBefore />}
+                {/* NOTE: StyledBefore is dont exist in InVision's mockup */}
+                {/* NOTE: ...see https://eurosport.invisionapp.com/d/main#/console/17267787/371330695/preview */}
+                {/* NOTE: index > 0 && <StyledBefore /> */}
+                {/* TODO: Find if this is deprecated... */}
```
davidgarry: You can remove it completely, we will not use it anymore

src/modules/PlayerCard/index.js
```diff
@@ -45,18 +46,19 @@ export const StyledScoreBlock = styled.div`
...
-    width: 19px;
-    ${breakpoints.medium(css`
-      width: 40px;
-    `)};
-    height: 20px;
-    border-right: 2px solid ${colors.coreLightMinus1};
-    display: block;
-  }
-`;
+// NOTE: Muted as not used... See L.126
```
JulienKode: Should we remove it as it is not used ?

src/modules/PlayerCard/mockData/playerCardMockData.js
```diff
@@ -166,3 +166,15 @@ export const previousMatches = [
     },
   },
 ];
+
+export const simpleLabels = [
```
JulienKode: Should we add this parameter into tests ?

src/modules/PlayerCard/index.js
degouville: Yes !

degouville:
Thanks @JulienKode 😉

degouville:
Thanks @Xb0X ;)

src/modules/PlayerCard/index.test.js
```diff
@@ -26,4 +26,7 @@ describe('PlayerCard', () => {
     const backgroundImg = wrapper.find(StyledBackground);
     expect(backgroundImg).toHaveStyleRule('background-image', `url(${BACKGROUND_IMG})`);
   });
+
+  // TODO: it should display previousMatches
+  // TODO: it should display labels in previousMatches
```
davidgarry: Maybe some tests are missing ?

degouville: Oh yes.

degouville: Done  ✅

PR #266, opened 2019-07-19
Ewl 827 table of rounds

narajaon:


src/modules/RoundTable/RoundTable.js
```diff
@@ -0,0 +1,97 @@
...
+const SingleDashPlaceholder = styled(Placeholder)`
+  ${({ bottom }) => bottom && 'border-bottom: none'};
+  border-right: none;
+  border-left: none;
+  border-top: none;
+`;
+
+const NeutralH5 = styled(H5)`
+  color: ${coreNeutral4};
+  ${fontAlphaHeadline};
```
JulienKode: Can we use colors from theme ?

src/modules/RoundTable/RoundTable.js
```diff
@@ -0,0 +1,97 @@
...
+    <FlexedContainer>
+      <Column>
+        {matches.map(match => (
+          <StyledScoreBlock key={match.id}>
+            <ScoreBlock.ScoreBlock matchUrl={match.matchUrl} data={match.data} hasButton={false} />
+          </StyledScoreBlock>
+        ))}
+      </Column>
+      <Column>
+        {matches.length > 1 && Array.from(Array(matches.length / 2)).map(() => <HalvedPlaceholder matches={matches} />)}
```
JulienKode: Should we add test for the case where we have one match ? 🤔

src/modules/RoundTable/RoundTable.js
```diff
@@ -0,0 +1,97 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'react-emotion';
+import { groupBy } from 'lodash';
+
+import { H5, fontAlphaHeadline } from '../../typography';
```
JulienKode: Should we use absolute imports ? 

What do you think @narajaon

src/modules/RoundTable/RoundTable.js
```diff
@@ -0,0 +1,97 @@
...
+
+const Column = styled(FlexedColum)`
+  height: 100%;
+  justify-content: space-around;
+`;
+
+const Placeholder = styled.div`
+  min-width: 30px;
+  height: 100%;
+  border: 2px solid rgba(79, 82, 106, 0.5);
```
JulienKode: Can we move that color to our colors file ?

narajaon: Good catch !

src/modules/RoundTable/RoundTable.js
narajaon: Good catch !

src/modules/RoundTable/RoundTable.js
narajaon: Good catch ! (i'll do another PR for all the theming / colors as soon as it's on master)

PR #270, opened 2019-07-23
Ewl 856 create rounds table mobile

degouville:


src/modules/RoundTable/RoundTableMobile.js
```diff
@@ -0,0 +1,86 @@
+import React, { Component } from 'react';
+import styled from 'react-emotion';
+import PropTypes from 'prop-types';
+
+import ScoreBlock, { scoreBlockType } from '../../components/ScoreBlock';
+import SimpleTabs from '../SimpleTabs';
+
+const StyledMatches = styled.div``;
+const StyledSimpleTabs = styled.div`
+  border-bottom: solid #fff4 1px;
```
JulienKode: Can we use this colors from the theme ?

src/modules/RoundTable/RoundTableMobile.js
```diff
@@ -0,0 +1,86 @@
+import React, { Component } from 'react';
+import styled from 'react-emotion';
+import PropTypes from 'prop-types';
+
+import ScoreBlock, { scoreBlockType } from '../../components/ScoreBlock';
```
JulienKode: Should we import theses files with an absolute path

src/modules/RoundTable/RoundTableMobile.js
```diff
@@ -0,0 +1,86 @@
...
+    initialTab: PropTypes.number,
+  };
+
+  static defaultProps = {
+    initialTab: 0,
+  };
+
+  componentDidMount() {
+    const { initialTab } = this.props;
+    this.setState({ currentTab: initialTab });
```
JulienKode: This component can be a good start to use `Hooks`, what do you think ? 🤔
But anyway it's not mandatory and works really well like that ✅

src/modules/RoundTable/RoundTableMobile.js
```diff
@@ -0,0 +1,86 @@
...
+    const { matches: rawMatches, tabs } = this.props;
+    const currentRound = [...tabs][currentTab].href || 1;
+    const isCurrentRound = number => ({ round }) => +number === round;
+    const matches = rawMatches.filter(isCurrentRound(currentRound));
+    const hasMatches = matches && !!matches.length;
+    const hasTabs = tabs && !!tabs.length;
+
+    return (
+      <>
+        {hasTabs && (
```
JulienKode: Can we add tests where `hasTabs` is `true` and another one for the `false` state ?

src/modules/RoundTable/RoundTableMobile.js
```diff
@@ -0,0 +1,86 @@
...
+          <StyledSimpleTabs>
+            <SimpleTabs initialTabIndex={currentTab} onItemSelected={this.changeTab} tabs={tabs} />
+          </StyledSimpleTabs>
+        )}
+
+        {hasMatches && (
```
JulienKode: Can we add tests where `hasMatches ` is `true` and another one for the `false` state ?

src/modules/RoundTable/mock/data.json
```diff
@@ -0,0 +1,7337 @@
+{
```
JulienKode: Where did you use this file ?

degouville: It not used. I'll remove it

src/theme.js
```diff
@@ -70,6 +71,14 @@ export default {
       fontFamily: fontInterUi,
     },
   },
+
+  tab: {
+    active: {
+      color: coreLightMinus1,
+      borderColor: featureOneMinus2,
+    },
+    borderColor: transparentize(0.5, coreLightMinus1),
```
JulienKode: Really good way to handle that @degouville ✅

src/modules/SimpleTabs/index.js
```diff
@@ -18,13 +17,13 @@ const StyledCarousel = styled(Carousel)`
   `)}
 `;
 
-export const TabCommonStyle = css`
+export const TabCommonStyle = ({ tab }) => css`
```
JulienKode: 🥇

src/modules/SimpleTabs/index.js
```diff
@@ -126,11 +127,16 @@ export default class SimpleTabs extends React.Component {
...
-        {tabs.map(({ label, disabled, href }, i) => {
+        {tabs.map(({ label, disabled, highligthed, href }, i) => {
           if (currentTab === i) return <TabSelected key={`${label}${href}`}>{label}</TabSelected>;
           if (disabled) return <TabInactive key={`${label}${href}`}>{label}</TabInactive>;
           return (
-            <TabActive key={`${label}${href}`} onClick={e => this.handleClick(e, href, i)} href={href}>
+            <TabActive
+              key={`${label}${href}`}
+              ishighligthed={highligthed}
+              onClick={e => this.handleClick(e, href, i)}
```
JulienKode: can we consider using a `useCallback` may be to avoid re-creating the function at each render ?

src/modules/LiveEventHero/index.js
```diff
@@ -14,6 +14,88 @@ const LiveEventHero = ({ title, subtitle }) => (
...
+  watchButtonText,
+  onWatchButtonClick,
+  highlightLastSet,
+}) => (
+  <StyledWrapper className={className}>
+    <Labels labels={labels} />
+    <StyledScoreWrapper>
+      <SetsScore data={scoreData} highlightLastSet={highlightLastSet} />
+    </StyledScoreWrapper>
+    {(courtInfo || date) && (
```
JulienKode: Should we add test for that ?
There are a lot of conditional rendering that can leads to many different view for the user

src/modules/SimpleTabs/index.js
JulienKode: Or you can do something like that: https://reactjs.org/docs/faq-functions.html#example-passing-params-using-data-attributes

JulienKode: so in the render it will be something like that: 

```
       <TabActive
               data-index={i}
               data-href={href}
               key={`${label}${href}`}
               ishighligthed={highligthed}
               onClick={e => this.handleClick(e, href, i)}
               href={href}
             >
```

and then in `handleClick`

```
  handleClick(e) {
    const tabIndex = e.target.dataset.index
    const href = e.target.dataset.href
    e && e.preventDefault();
    const { onItemSelected } = this.props;

    onItemSelected(href, tabIndex);
    this.setState({ selectedTabIndex: tabIndex });
  }
```

JulienKode: What do you think @degouville ?

PR #272, opened 2019-07-23
(EWL-915): Video Player Modal for Login integration 📱🖥

JulienKode:
# [(EWL-915)](https://eurosportdigital.atlassian.net/browse/EWL-915): Video Player Modal for Login integration 📱🖥

## Checklist

- Browser compatibility checked:
  - [x] Safari
  - [x] Chrome
  - [x] Firefox
- [x] MR title respects the format: `(EWL-XXX): description`

### Modal

|         Mockup         |      New component     |  
| :--------------------: | :--------------------: | 
| Screenshot 2019-07-23 at 14 41 21 |  Screenshot 2019-07-23 at 14 41 41 | 

## Description

you can test the `VideoPlayerModal` component here: https://deploy-preview-272--es-web-toolkit.netlify.com

- [x] create `VideoPlayerModal` component
- [x] handle browser and mobile devices

## Test Plan

- create a stories on storybook for `VideoPlayerModal`
- use it with login
- test mobile and browser with responsive


src/modules/Login/index.stories.js
```diff
@@ -3,7 +3,8 @@ import { storiesOf } from '@storybook/react';
 import { text } from '@storybook/addon-knobs';
 import Login from './index';
 
-const createDefaultProps = () => ({
+// eslint-disable-next-line import/prefer-default-export
```
saad-kh: I'm not a fan of using `// eslint-disable-next-line`... Either the rule is too opinionated and we remove it, or we find a way to avoid triggering it.

JulienKode: Yes but it does not have to be a default export for that case as it's not the main purpose of that file

JulienKode:
the login component is different from the modal component, in this modal component we will add overlays and design of the modal which is different from the login

src/modules/Login/index.stories.js
JulienKode: Can we ignore `stories` files ?

saad-kh: * Using a placeholder instead of the actual `Login` component in `src/modules/VideoPlayerModal/index.stories.js`
* Giving plain props to this `Login` component instead of relying on an external helper function
* Making sure the `Login` component behaves well with minimal props that don't lead to the creation of a helper function
* Removing the eslint rule

saad-kh:
Ah I okay I see

src/modules/Login/index.stories.js
JulienKode: Makes sense 

d8957680874228eedfc5efd609f14456cc52916c should solve the issue

PR #291, opened 2019-07-31
Fix edge issue on 8 digits hexa colors

leozero:
# About the type of changes

If there's no conflicts
- [ ] squash and merge
- [ ] delete branch


src/modules/SubNavigation/index.js
```diff
@@ -12,7 +13,7 @@ const StyledArrowCSS = css`
   height: 100%;
   background: transparent;
   padding: 0 6px;
-  border-left: solid 1px ${loftySilver};
+  border-left: solid 1px ${rgba(loftySilver, 0.1)};
```
JulienKode: Should we use https://polished.js.org/docs/#transparentize ?

PR #288, opened 2019-07-29
(EWL-948): Player bottom bar and controls

JulienKode:
# (EWL-948): Player bottom bar

## Checklist

- Browser compatibility checked:
  - [x] Safari
  - [x] Chrome
  - [x] Firefox
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Design

|         Mockup         |      Large device      |     
| :--------------------: | :--------------------: |
| Screenshot 2019-07-31 at 17 35 01 |   |

|         Mockup         |      Small device      |    
| :--------------------: | :--------------------: |
| Screenshot 2019-07-31 at 17 38 15 |  |

## Demo

demo

## Description

This PR is the next of https://github.com/EurosportDigital/web-toolkit/pull/287

- [x] add bottom bar design ✅
- [x] add live button ✅
- [x] add rewind counts when we are not in Live ✅
- [x] add fullscreen button ✅
- [x] add controls skin ✅
- [x] make different condition to render the skin on mobile and desktop ✅
- [x] add player buttons ✅
- [x] mix https://github.com/EurosportDigital/web-toolkit/pull/287 and https://github.com/EurosportDigital/web-toolkit/pull/288

Related to: https://github.com/EurosportDigital/eurosport-toolkit-player/pull/55

## Future implementation

Theses features will be added in other PR's:

- [x] Video time slider
- [x] Volume
- [x] Captions
- [x] Fullscreen plugged to JWPlayer

# Checklist
### Before code review

- [ ] Move Jira ticket into PR
- [x] Check with another developers that features sounds good

### After code review

- [ ] Move Jira ticket into QA

PR #294, opened 2019-07-31
(Fix): RoundTable tests

JulienKode:
# About the type of changes

Seems that we have problem here: https://github.com/EurosportDigital/web-toolkit/blob/e854ff446699a50ccbdeb605d18be1fb5e3db046/src/modules/RoundTable/__snapshots__/index.test.js.snap#L5475

We made a snapshot of the whole theme,
so each time we change the theme, even if it's not related we will have to update this test...

Here is what we currently have
```
    const wrapper = mount(
      <ThemeProvider theme={theme}>
        <RoundTableMobile matches={matches} tabs={tabs} />
      </ThemeProvider>
    );
    expect(wrapper).toMatchSnapshot();
    wrapper.unmount();
```

Here is what I suggest:

```
    // Given
    const wrapper = mount(
      <ThemeProvider theme={theme}>
        <RoundTableMobile matches={matches} tabs={tabs} />
      </ThemeProvider>
    );
    
    // When
    const roundTable = wrapper.find(RoundTableMobile)
    
    // Expect
    expect(roundTable).toMatchSnapshot();
    wrapper.unmount();
```

https://github.com/EurosportDigital/web-toolkit/pull/294 will fix the issue

see: https://github.com/EurosportDigital/web-toolkit/pull/292#issuecomment-516786741


JulienKode:
we also had the problem for Dropdown it should be fixed by https://github.com/EurosportDigital/web-toolkit/pull/288/commits/00eaa9b66e726124eb098fa693c25fdd9f31079f

PR #304, opened 2019-08-06
Ewl 824 add initial choice to props

degouville:
# About the type of changes

# Test plan

- [ ] Empty data
- [ ] Invalid data
- [X] Correct data


src/components/Dropdown/index.js
```diff
@@ -211,15 +215,18 @@ class Dropdown extends React.Component {
 
 Dropdown.defaultProps = {
   onItemSelected: null,
+  initialOptionID: null,
 };
 
+// TODO: Add initial choice
```
davidgarry: TODO ?

src/components/Dropdown/index.js
```diff
@@ -156,8 +156,12 @@ class Dropdown extends React.Component {
   }
 
   setSelectedOption = () => {
-    const { options } = this.props;
-    this.setState({ selectedOption: options[0] });
+    const { options, initialOptionID } = this.props;
+    const matchInitialID = initialID => ({ id }) => id === initialID;
+    const initialOption = options.find(matchInitialID(initialOptionID));
+    const firstOption = options[0];
+    const selectedOption = initialOption || firstOption;
```
davidgarry: I think you can reduce the number of lines for that :
```
const initialOption = options.find(({ id }) => id === initialOptionID);
const selectedOption = initialOption || options[0];
```

src/components/Dropdown/index.js
ra-phael: Can't this comment be removed now that initialOptionID was added?

src/components/Dropdown/index.stories.js
```diff
@@ -2,8 +2,15 @@ import React from 'react';
...
 
 const dropdownStories = storiesOf('Components|Dropdown', module);
 
-dropdownStories.add('Default', () => <Dropdown options={object('options', dropdownOptions)} />);
+dropdownStories
+  .add('Default', () => <Dropdown options={object('options', dropdownOptions)} />)
+  .add('With Initial Option', () => (
+    <Dropdown
+      options={object('options', dropdownOptions)}
+      initialOptionID={object('InitialOptionID', initialOptionID)}
```
davidgarry: This one is not an object knob

ra-phael: Looks like it's a number, not an object 🤔

src/components/Dropdown/index.js
degouville: yes

src/components/Dropdown/index.js
degouville: ok

src/components/Dropdown/index.stories.js
degouville: Right.

FIXED ✅

src/components/Dropdown/index.js
degouville: REMOVED ✅

src/components/Dropdown/index.stories.js
degouville: Right.

FIXED ✅

src/components/Dropdown/index.js
```diff
@@ -156,8 +156,11 @@ class Dropdown extends React.Component {
   }
 
   setSelectedOption = () => {
-    const { options } = this.props;
-    this.setState({ selectedOption: options[0] });
+    const { options, initialOptionID } = this.props;
+    const matchInitialID = initialID => ({ id }) => id === initialID;
```
JulienKode: Should we test an option change ?

degouville: I've made a test case for initial setting [here](https://github.com/EurosportDigital/web-toolkit/blob/06db19856937b16791d354db28b6f545755282c3/src/components/Dropdown/index.test.js#L50)

By "option change" you mean to test *L172* like this ? :
```
  it('should display selected option after new selection', () => {
    const wrapper = ...

    wrapper.find(StyledLi).get(2).simulate('click');

    const isSelected = li => !!li.prop('isSelected');
    const selectedOption = wrapper.findWhere(isSelected);
    const selectedOptionText = selectedOption.text();
    expect(selectedOptionText).toBe('Men doubles ✓');
    wrapper.unmount();
  });
```

JulienKode: Oh perfect I've just see that `setSelectedOption` is call on mount

degouville: Maybe "setSelectedOption" should be "setInitalOption" as there is a confusion... What do you think ?

JulienKode: @degouville good idea :)

degouville: Done ✅

PR #309, opened 2019-08-08
(EWL-970): Slider component

JulienKode:
# (EWL-970): Slider component

## About the type of changes

###

|        **Horizontal**         |      **Small**      |      **Large**      |
| :--------------------: | :--------------------: | :--------------------: |
| MockupHorizontal | small-horizontal | large-horizontal |
|         **Vertical**        |      **Small**      |     **Large**      |
| Screenshot 2019-08-08 at 14 19 30 | small-vertical | large-vertical |

## Description

 We use [`rc-slider`](http://github.com/react-component/slider/) library instead of `<input type="range" />`  that we custom with css because each browser handle custom input range with different css properties that cannot be autoprefixed with our current solution as example to show a color of progress, Mozilla handle that, but not chrome, safari handle it in a different way and Microsoft explorer too.

- [x] add identity-obj-proxy from https://jestjs.io/docs/en/webpack to mock css modules inside our snapshots
- [x] create slider and handle vertical and horizontal behaviour
- [x] use rc-slider library instead of `<input type="range" />`  that we custom with css
- [x] create different slider theme
- [x] use slider with a theme

## Test

- tested with horizontal behaviour
- tested with vertical behaviour

## Test plan

- start using it in vertical 
- start using it in horizontal 

### Before code review


src/components/JWPlayerETP/Controls/Slider/index.js
```diff
@@ -0,0 +1,63 @@
...
+  backgroundColor: handleColor,
+  margin: -(handleSize / 2) + railThickness / 2,
+});
+
+const getBarStyle = ({ backgroundColor, railThickness, vertical }) => {
+  if (vertical) return { backgroundColor, width: railThickness };
+  return { backgroundColor, height: railThickness };
+};
+
+const SliderComponent = ({
```
degouville: Why this component is named... Slider**Component** ?

Maybe could you `import Slider as RcSlider from 'rc-slider';` and then just name ours `<Slider />` ?

JulienKode: Good idea @degouville 😊

JulienKode: d43c1365950be103d8d266706ecc7f64a4e547f1 should solve the issue

PR #311, opened 2019-08-09
(EWL-989): Seek component

JulienKode:
# About the type of changes

# Design 

|         VOD         |     Live      |     
| :--------------------: | :--------------------: | 
| vod | live-video |

# Description

- [x] create seek component from values of ETP https://github.com/EurosportDigital/eurosport-toolkit-player/pull/67
- [x] keep the same width based on characters for rewind counts 

# Test plan

- tested with the version of ETP from https://github.com/EurosportDigital/eurosport-toolkit-player/pull/67


src/components/JWPlayerETP/Controls/BottomBar/index.test.js
```diff
@@ -9,6 +9,18 @@ describe('Components|JWPlayerETP|Controls|BottomBar', () => {
     isLive: true,
     children: <Component />,
     rewindCounts: undefined,
+    theme: {
```
narajaon: Maybe importing the real theme (if there's any) and adding it here would be better ?

narajaon: from /src/theme.js i mean

JulienKode: Here we want to mock it and not being dependent from it because if the theme change, we will have to update test (the snapshot here) which is something that we do not want I think

PR #313, opened 2019-08-12
Ewl 536 add schedule to score block

degouville:
# About the type of changes

# Test plan

- [ ] Empty data
- [ ] Invalid data
- [X] Correct data

# Checklist
### Before code review

- [X] Move Jira ticket into PR
- [X] Check with another developers that features sounds good

### After code review

### After QA validation
  - [ ] Fix conflicts

If there's no conflicts
- [ ] squash and merge
- [ ] delete branch


src/components/ScoreBlock/ScoreBlock.js
```diff
@@ -180,6 +185,8 @@ ScoreBlock.defaultProps = {
   isLive: false,
   // eslint-disable-next-line react/default-props-match-prop-types
   isWatchable: false,
+  // eslint-disable-next-line react/default-props-match-prop-types
```
narajaon: Do we have to put this eslint comment ?

degouville: Yes.

Seems like this is required if we use spread operator on [L197](https://github.com/EurosportDigital/web-toolkit/blob/11aeab8f090add5153c94d67fc2376fd2e1acd3f/src/components/ScoreBlock/ScoreBlock.js#L197)

narajaon: Alright

src/components/ScoreBlock/ScoreBlock.js
```diff
@@ -167,8 +171,10 @@ export const ScoreBlock = ({
 export const scoreBlockType = {
   matchUrl: PropTypes.string.isRequired,
   data: setsScoreType.isRequired,
+  schedule: setsScheduleType.isRequired,
```
davidgarry: Maybe we can put this prop optional and remove the isScheduled prop.
If schedule is null, it means we don't want to display the schedule.

src/components/ScoreBlock/SetsSchedule.js
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'react-emotion';
+import * as breakpoints from '../../breakpoints';
+
+export const StyledSetsSchedule = styled.p`
+  padding: 0 16px;
```
davidgarry: If the text is long, it's wrapped on two lines. Can you change that to always keep one line ?
image

src/components/ScoreBlock/SetsSchedule.test.js
```diff
@@ -0,0 +1,40 @@
...
+        <SetsSchedule data={players} schedule={schedule} />
+      </ThemeProvider>
+    );
+  });
+
+  afterEach(() => {
+    wrapper.unmount();
+  });
+
+  it('Renders a SetsSchedule component', () => {
```
davidgarry: I think for this simple test, you can use only a toMatchSnapshot. It will test directly that all components are set ;)

src/components/ScoreBlock/SetsSchedule.js
JulienKode: Should we use a size with `ch` also ?

src/components/ScoreBlock/ScoreBlock.js
degouville: Ok i see what you mean.

I made this choice to use `isScheduled` prop so we can toggle it in storybook and as we could choose to display it or not EVEN if we dont have _Sets_ setted.

I'll set `schedule: setsScheduleType` tho.
So maybe we can keep `hasSchedule` filter AND `isSchedule` prop too ?

What do you think ?

src/components/ScoreBlock/SetsSchedule.js
degouville: OK

src/components/ScoreBlock/SetsSchedule.test.js
degouville: LESS IS MORE ✅

JulienKode: it makes sense as `StyledSetsSchedule` will be in the snapshot

degouville: Right

src/components/ScoreBlock/SetsSchedule.js
degouville: @JulienKode `flex: none` do the job

src/components/ScoreBlock/ScoreBlock.js
degouville: isSchedule removed

degouville:
Thanks U guys for this review.

PR #337, opened 2019-08-21
pass match props to scoreblock

degouville:
# About the type of changes

# Test plan

- [ ] Empty data
- [ ] Invalid data
- [x] Correct data


src/modules/RoundTable/RoundTableMobile.js
```diff
@@ -72,9 +72,9 @@ export default class RoundTableMobile extends Component {
 
         {hasMatches && (
           <StyledMatches>
-            {matches.map(({ id, matchUrl, data }) => (
+            {matches.map(({ id, ...match }) => (
```
narajaon: I prefere
```
matches.map(match => (
<StyledScoreBlock key={match.id}>
   <ScoreBlock.ScoreBlock {...match} hasButton={false} />
...
```
It's less computation :)

degouville: I prefer this early destructuring as it show us we will need `id` a `data` as variable. 😅🧐
also `key={Z}`seems more readable than `key={match.X.Y.Z}`

what do you thing ?

narajaon: I agree with destructuring as a way to avoid too much chaining. We should just be careful not to overuse the spread operator

degouville: All right

PR #338, opened 2019-08-21
(EWL-1030): Change volume behaviour

JulienKode:
# About the type of changes

# Design

|         Mockup         |      Before      |     Now      |
| :--------------------: | :--------------------: | :--------------------: |
| mockup | volume | newVolume |

# Description

Change the way of using the volume for web and mobile.

## Before

On web and mobile:
- it was needed to click on the volume to show the dropdown 
- then you can change the volume value
- when the volume in the slider is 0, it is considered as muted

## Now

On mobile:
- just a volume and mute button 
- you can mute by just tapping on the volume button
- cannot change the volume from JWPlayer, control the volume from you mobile device

On web:
- just a volume and mute button 
- you can mute by just tapping on the volume button
- when the mouse is over the volume section you can change the volume of the component 
- when you unmute it use the last volume that you set

# Test Plan

- [x] tested with web-toolkit https://github.com/EurosportDigital/eurosport-toolkit-player/pull/76
- [x] see that we can mute / unmute by just clicking
- [x] see that we can change the volume when we go to 0 it sets to mute
- [x] when we mute / unmute we came back to the last volume used 

# Checklist
### Before code review

### After code review


src/components/JWPlayerETP/Controls/Volume/useVolumeWithDelay.js
```diff
@@ -0,0 +1,19 @@
...
+  const [showVolume, setShowVolume] = useState(initialState);
+  const [willShowVolume, setWillShowVolume] = useState(initialState);
+
+  useEffect(() => {
+    if (willShowVolume === false) {
+      const timer = setTimeout(() => setShowVolume(false), 200);
+      return () => clearTimeout(timer);
+    }
+    setShowVolume(true);
+    return undefined;
```
narajaon: Why return undefined here ? To have consistent returns I guess ?
Wouldn't it be cleaner to just remove it ?

JulienKode: It's for the linter, but you're right without that it does the same

PR #349, opened 2019-08-23
(Ewl-1041): centralize proptypes

narajaon:
# About the type of changes

# Goal

Centralize types to avoid repeating / useless type-checking with the increase of smart components on ETS.


narajaon:
@degouville great point, as discussed earlier I will instead remove the type suffix from scoreBlock to unify everything

src/modules/RoundTable/RoundTableMobile.js
```diff
@@ -26,21 +27,8 @@ export default class RoundTableMobile extends Component {
   };
 
   static propTypes = {
-    matches: PropTypes.arrayOf(
-      PropTypes.shape({
-        ...scoreBlockType,
-        hasWon: PropTypes.bool,
-        round: PropTypes.object,
```
davidgarry: Props hasWon and round has been deleted
- hasWon should not be here
- round is used

PR #358, opened 2019-08-29
(Ewl-1068): fix styledhr

narajaon:
# About the type of changes


src/modules/RoundTable/RoundTableMobile.js
```diff
@@ -14,7 +15,7 @@ const StyledScoreBlock = styled.div`
   margin: 2em 0 2em 0;
 `;
 const StyledHr = styled.div`
-  background-color: ${({ theme }) => theme.tab.borderColor};
+  background-color: ${coreNeutral9};
```
JulienKode: Can we get it from the theme ?

PR #361, opened 2019-08-30
EWL-1013 improve viewmore height refresh

degouville:
# About the type of changes

# Test plan

- [ ] Empty data
- [ ] Invalid data
- [ ] Correct data

# Checklist
### Before code review

- [ ] Move Jira ticket into PR
- [ ] Check with another developers that features sounds good

### After code review

### After QA validation
  - [ ] Fix conflicts

If there's no conflicts
- [ ] squash and merge
- [ ] delete branch


src/modules/ViewMore/debug.stories.js
```diff
@@ -0,0 +1,157 @@
+// TODO: REMOVE THIS AFTER REVIEW
```
degouville: Note that this file is for reproduce the bug only.

It will be deleted after review and not published on the production storybook

src/modules/ViewMore/index.js
```diff
@@ -56,46 +61,71 @@ export default class ViewMore extends React.Component {
...
   }
 
-  computeBlockHeight() {
+  computeBlockHeight(height) {
     const { showLessItemCount } = this.props;
     const { expanded } = this.state;
     const listElement = get(this, 'listRef.current', {});
     const listChildren = get(listElement, 'children', []);
+    const shouldExpand = expanded || showLessItemCount >= listChildren?.length;
+    const expandedSize = height;
```
narajaon: why not just name height expendedSize then ?

src/modules/ViewMore/index.js
```diff
@@ -25,19 +25,24 @@ export const StyledViewMoreButton = styled.span`
...
+      overflow: ${isDelayGone ? 'unset' : 'hidden'};
+
+      // NOTE: <SizeMe> required to have an extra height to render properly
+      border-top: 1em solid transparent;
+      margin-top: -1em;
     `}
 `;
 
-export default class ViewMore extends React.Component {
+class ViewMore extends React.Component {
```
narajaon: Why not use a functional component if you had to rewrite everything ?

degouville: I've decided precisely to keep the already working ViewMore and just improve it.

src/modules/ViewMore/index.js
degouville: I personally think it better to understand that computeBlockHeight get a Height from sizeMe and then we set the expandedSize from it. 

But yes we could also computeBlockHeight(expendedSize) as there is no other calculation to set the expendedSize (from the moment).

I'm comfortable with the current declaration tho...
What do you think ?

src/modules/ViewMore/index.js
degouville: My first try to rewrite this component from scratch wasn’t better according to the design product requirements 😬.

PR #264, opened 2019-07-19
(EWL-897): New login modal UI - Mobile / Desktop / Tablet 🖥📱

JulienKode:
# [(EWL-897)](https://eurosportdigital.atlassian.net/browse/EWL-897): New login modal UI - Mobile / Desktop / Tablet 🖥📱

Please check the type of changes you're about to make

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Checklist

- Browser compatibility checked:
  - [x] Safari
  - [x] Chrome
  - [x] Firefox
- [x] MR title respects the format: `(EWL-XXX): description`
- [x] Move Jira ticket into PR

### Login

|         Mockup         |      New component     |  
| :--------------------: | :--------------------: | 
| Mockup | Login | 

## Example

|         Login         |   
| :--------------------: | 
| example |

## Description

you can test the Login component here: https://deploy-preview-263--es-web-toolkit.netlify.com

- [x] create Login component
- [x] handle submit by forwarding user credentials
- [x] handle update of credentials in the input
- [x] create other UI component 
- [x] create `UnderlineLink` component
- [x] add snapshot testing for `Login` component
- [x] add snapshot testing for `UnderlineLink` component
- [x] test expected behaviour of the `Login` component
- [x] use `Input` from #253 
- [x] use `ErrorBanner` from #255   
- [x] use `Button` with type form from #257  
- [x] use `ArrowLink` from #37 
- [x] use `UnderlineLink` from #264 

## Test Plan

- setup a desktop stories
- create a mobile stories 
- use them both with responsive on chrome, firefox and safari

PR #378, opened 2019-09-09
EWL-1079 integrate recaptcha

ra-phael:
# About the type of changes

# Test plan

Will need to be done with Cypress

# Checklist
### Before code review

- [ ] Check with another developers that features sounds good

### After code review

### After QA validation
  - [ ] Fix conflicts

If there's no conflicts
- [ ] squash and merge
- [ ] delete branch


src/modules/Login/index.js
```diff
@@ -25,22 +26,52 @@ const Login = ({
...
     event => {
       event.preventDefault();
-      onSubmit({ email, password });
+      isRecaptchaLoaded && recaptchaRef.current.execute();
     },
-    [email, onSubmit, password]
+    [recaptchaRef, isRecaptchaLoaded]
   );
 
+  const onRecaptchaChange = token => {
```
JulienKode: Should we test that function ?

src/modules/Login/index.js
```diff
@@ -25,22 +26,52 @@ const Login = ({
...
+    setRecaptchaToken(token);
+    if (token !== null) {
+      onSubmit({ email: emailRef.current, password: passwordRef.current, recaptchaToken: token });
+    } else {
+      recaptchaRef.current.execute();
+    }
+  };
+
+  const asyncRecaptchaOnLoad = useCallback(() => {
```
JulienKode: Should we test that hooks ?

src/modules/Login/index.js
```diff
@@ -25,22 +26,52 @@ const Login = ({
...
+  useEffect(() => {
+    emailRef.current = email;
+    passwordRef.current = password;
+  });
+
+  const [isRecaptchaLoaded, setisRecaptchaLoaded] = useState(false);
+  const [recaptchaToken, setRecaptchaToken] = useState(null);
+
+  const onSignInSubmit = useCallback(() => {
+    if (recaptchaToken !== null) {
```
JulienKode: Can we test that path ?

src/modules/Login/index.test.js
```diff
@@ -81,61 +80,6 @@ describe('Login', () => {
       expect(elements.length).toBe(1);
     });
 
-    it('Should trigger onSubmit when we click on submit button', () => {
```
JulienKode: Why removing this expected specs ? 
Do we need it to track regression 🤔

src/modules/Login/index.js
```diff
@@ -51,12 +82,12 @@ const Login = ({
...
       <InputsContainer>
         <ComponentContainer>
-          <Input value={email} onChange={onEmailChange} placeholder={emailPlaceholder} />
+          <Input value={email} type="text" onChange={e => setEmail(e.target.value)} placeholder={emailPlaceholder} />
         </ComponentContainer>
         <ComponentContainer>
           <Input
             value={password}
-            onChange={onPasswordChange}
+            onChange={e => setPassword(e.target.value)}
```
JulienKode: Can we use a `useCallback` ?

src/modules/Login/index.js
```diff
@@ -51,12 +82,12 @@ const Login = ({
       )}
       <InputsContainer>
         <ComponentContainer>
-          <Input value={email} onChange={onEmailChange} placeholder={emailPlaceholder} />
+          <Input value={email} type="text" onChange={e => setEmail(e.target.value)} placeholder={emailPlaceholder} />
```
JulienKode: Can we use a `useCallback` here too ?
Then we will not have to recreate a function at each render

src/modules/Login/index.js
JulienKode: Can we use that function with `useMemo` or `useCallback` ?
Then we will not have to re-create it at each render

src/modules/Login/index.js
davidgarry: Keep only one method (onSignInSubmit / onFormSubmit)

src/modules/Login/index.js
```diff
@@ -25,22 +26,52 @@ const Login = ({
   errorMessage,
   showSubscribeSection,
   isSecondaryMode,
+  recaptchaSiteKey,
 }) => {
+  const recaptchaRef = useRef(null);
+  const emailRef = useRef(null);
+  const passwordRef = useRef(null);
+
```
JulienKode: Should we move that part to a custom hooks ?

src/modules/Login/index.js
```diff
@@ -25,22 +26,52 @@ const Login = ({
...
     },
-    [email, onSubmit, password]
+    [recaptchaRef, isRecaptchaLoaded]
   );
 
+  const onRecaptchaChange = token => {
+    setRecaptchaToken(token);
+    if (token !== null) {
+      onSubmit({ email: emailRef.current, password: passwordRef.current, recaptchaToken: token });
+    } else {
```
davidgarry: You can remove the else because we always generate a token before the submission

src/modules/Login/index.js
davidgarry: type="text" ?

src/modules/Login/index.js
JulienKode: I agree it will be easier to maintain 

Just for references if you want to check the initial thinking about that implementation  https://github.com/EurosportDigital/web-toolkit/pull/264#discussion_r305348114

PR #421, opened 2019-11-19
fix(RPW2-26): import css from emotion

JulienKode:
# About the type of changes

# Checklist
### Before code review

### Description

with @ldestin @aminemoussaid @Xb0X 

- When we `import { css } from "emotion";` css class where missing inside the page
- So we imported it with `import { css } from "react-emotion";`

It was hard to find that the root cause was the import, in their documentation we can import css from [emotion](https://5bb1495273f2cf57a2cf39cc--emotion.netlify.com/docs/css)


src/modules/ViewMore/debug.stories.js
```diff
@@ -2,8 +2,8 @@
 
 import React, { useState, useEffect, useRef } from 'react';
 import { storiesOf } from '@storybook/react';
-import styled from 'react-emotion';
-import { css } from 'emotion';
+import styled, { css } from 'react-emotion';
+
 // eslint-disable-next-line import/no-extraneous-dependencies
 import { Global } from '@emotion/core';
```
thibaultboursier: I know it's not part of your changes, but do you think it could lead to some problems to use @emotion/core (version 10.*) with emotion 9 package ?

JulienKode: It will not cause problem as we will have to refactor it anyway.
- `react-emotion` has been deprecated since the version 10.0.0 https://www.npmjs.com/package/react-emotion we will have to use `@emotion/styled`. 

For the moment we have to use `react-emotion` as we are using the version `9.0.0`. Moreover you embrace a good subject as we have to make a migration to the version `10.0.0`. It is important as `9.0.0` is very old now and not supported anymore

thibaultboursier: OK, it's clear ;-)

